PR #6, opened 2019-07-31
[#167569704] Add GetLimitedProfile function

francescopersico:
Draft because we need to decide how to handle the routes of the two http triggered functions.

For GetLimitedProfile i have set `v1/profiles/{fiscalCode}`. If it is ok for you i can change the CreateMessage route from `{*segments}` to `v1/messages/{fiscalCode}`.

### DIFF
--- a/GetLimitedProfile/__tests__/handler.test.ts
+++ b/GetLimitedProfile/__tests__/handler.test.ts
@@ -22,7 +22,7 @@ import {
 } from "../handler";
 
 describe("isSenderAllowed", () => {
-  it("should return false is the service is not allowed to send notifications to the user", () => {
+  it("should return false if the service is not allowed to send notifications to the user", () => {
     const blockedInboxOrChannels: IProfileBlockedInboxOrChannels = {
       "01234567890": new Set([BlockedInboxOrChannelEnum.INBOX])
     };
@@ -35,7 +35,7 @@ describe("isSenderAllowed", () => {
     expect(isAllowed).toBe(false);
   });
 
-  it("should return true is the service is allowed to send notifications to the user", () => {
+  it("should return true if the service is allowed to send notifications to the user", () => {
     const blockedInboxOrChannels: IProfileBlockedInboxOrChannels = {};
 
     const isAllowed = isSenderAllowed(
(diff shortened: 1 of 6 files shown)


GetLimitedProfile/handler.ts
```diff
@@ -0,0 +1,154 @@
...
+ */
+export function GetLimitedProfile(
+  serviceModel: ServiceModel,
+  profileModel: ProfileModel
+): express.RequestHandler {
+  const handler = GetLimitedProfileHandler(profileModel);
+
+  const middlewaresWrap = withRequestMiddlewares(
+    AzureApiAuthMiddleware(
+      new Set([UserGroup.ApiLimitedProfileRead, UserGroup.ApiFullProfileRead])
```
cloudify: since this API will be used only by 3rd party services, we can limit the groups to `ApiLimitedProfileRead` (`ApiFullProfileRead` is only used by the app backend)

```suggestion
      new Set([UserGroup.ApiLimitedProfileRead])
```

GetLimitedProfile/__tests__/handler.test.ts
```diff
@@ -0,0 +1,183 @@
...
+  profileArb
+} from "../../utils/arbitraries";
+import {
+  GetLimitedProfileHandler,
+  isSenderAllowed,
+  toLimitedProfile
+} from "../handler";
+
+describe("isSenderAllowed", () => {
+  it("should return false is the service is not allowed to send notifications to the user", () => {
```
cloudify:
```suggestion
  it("should return false if the service is not allowed to send notifications to the user", () => {
```

GetLimitedProfile/__tests__/handler.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    const isAllowed = isSenderAllowed(
+      blockedInboxOrChannels,
+      "01234567890" as NonEmptyString
+    );
+
+    expect(isAllowed).toBe(false);
+  });
+
+  it("should return true is the service is allowed to send notifications to the user", () => {
```
cloudify:
```suggestion
  it("should return true if the service is allowed to send notifications to the user", () => {
```

cloudify:
> Draft because we need to decide how to handle the routes of the two http triggered functions.
> 
> For GetLimitedProfile i have set `v1/profiles/{fiscalCode}`. If it is ok for you i can change the CreateMessage route from `{*segments}` to `v1/messages/{fiscalCode}`.

yes go ahead, thanks

digitalcitizenship:
## Affected stories

  * 🌟 [#167569704](https://www.pivotaltracker.com/story/show/167569704): Creare in io-functions-services la function GetLimitedProfile

<p align="right">
</p>

PR #21, opened 2019-09-23
[#168694799] Update specification for CreateMessage

francescopersico:
Updated with changes from #18

### DIFF
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -98,9 +98,66 @@ schemes:
 security:
 - SubscriptionKey: []
 paths:
-  "/messages/{fiscal_code}":
+  "/messages":
     post:
       operationId: submitMessageforUser
+      summary: Submit a Message
+      description: |-
+        Submits a message to a user.
+        On error, the reason is returned in the response payload.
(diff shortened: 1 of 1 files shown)


openapi/index.yaml
```diff
@@ -110,6 +110,63 @@ paths:
...
+          description: Unauthorized
+        '429':
+          description: Too many requests
+        '500':
+          description: The message cannot be delivered.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+  "/messages/{fiscal_code}":
+    post:
+      operationId: submitMessageforUserWithFiscalCode
```
cloudify: this breaks backward compatibility, this operation name should not change (`submitMessageforUser`), the other one (new) should have a new name

digitalcitizenship:
## Affected stories

  * 🌟 [#168694799](https://www.pivotaltracker.com/story/show/168694799): Modificare il file delle specifiche per la funzione CreateMessage

<p align="right">
</p>

openapi/index.yaml
francescopersico: Exactly the reason why i asked if the operationId was used by the client generator. I Have fixed with the new commit.

PR #25, opened 2019-11-07
[#169123489] add test coverage to GetSubscriptionsFeed

giux78:
Added test to GetSubscriptionsFeed still draft waiting for comments @cloudify @francescopersico

### DIFF
--- a/GetSubscriptionsFeed/__tests__/handler.test.ts
+++ b/GetSubscriptionsFeed/__tests__/handler.test.ts
@@ -1,0 +1,121 @@
+jest.mock("winston");
+
+import { TableService } from "azure-storage";
+import { MaxAllowedPaymentAmount } from "io-functions-commons/dist/generated/definitions/MaxAllowedPaymentAmount";
+import { toAuthorizedCIDRs } from "io-functions-commons/dist/src/models/service";
+import {
+  IAzureApiAuthorization,
+  UserGroup
+} from "io-functions-commons/dist/src/utils/middlewares/azure_api_auth";
+import { IAzureUserAttributes } from "io-functions-commons/dist/src/utils/middlewares/azure_user_attributes";
+import { NonNegativeNumber } from "italia-ts-commons/lib/numbers";
+import {
+  EmailString,
+  NonEmptyString,
+  OrganizationFiscalCode
+} from "italia-ts-commons/lib/strings";
+
+import { GetSubscriptionsFeedHandler } from "../handler";
+import { TableEntry } from "../utils";
+
+afterEach(() => {
+  jest.resetAllMocks();
+  jest.restoreAllMocks();
+});
+
+const anOrganizationFiscalCode = "01234567890" as OrganizationFiscalCode;
+const anEmail = "test@example.com" as EmailString;
+
+const aUserAuthenticationDeveloper: IAzureApiAuthorization = {
+  groups: new Set([UserGroup.ApiMessageRead, UserGroup.ApiMessageWrite]),
+  kind: "IAzureApiAuthorization",
+  subscriptionId: "s123" as NonEmptyString,
+  userId: "u123" as NonEmptyString
+};
+
+const someUserAttributes: IAzureUserAttributes = {
+  email: anEmail,
+  kind: "IAzureUserAttributes",
+  service: {
+    authorizedCIDRs: toAuthorizedCIDRs([]),
+    authorizedRecipients: new Set([]),
+    departmentName: "IT" as NonEmptyString,
+    isVisible: true,
+    maxAllowedPaymentAmount: 0 as MaxAllowedPaymentAmount,
+    organizationFiscalCode: anOrganizationFiscalCode,
+    organizationName: "AgID" as NonEmptyString,
+    requireSecureChannels: false,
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * ⚙️ [#169123489](https://www.pivotaltracker.com/story/show/169123489): Aumentare test coverage di io-functions-services

<p align="right">
</p>

GetSubscriptionsFeed/utils.ts
```diff
@@ -7,7 +7,7 @@ import { FiscalCodeHash } from "../generated/definitions/FiscalCodeHash";
 /**
  * A minimal storage table Entry
  */
-type TableEntry = Readonly<{
+export type TableEntry = Readonly<{
```
gunzip: there's an open PR to add these types into io-functions-commons here:  https://github.com/teamdigitale/io-functions-commons/pull/31/files#diff-6acba14f29206bf08111645a46659d1fR233

cc @francescopersico

GetSubscriptionsFeed/__tests__/handler.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { GetSubscriptionsFeedHandler } from "../handler";
+import { TableEntry } from "../utils";
+
+afterEach(() => {
+  jest.resetAllMocks();
+  jest.restoreAllMocks();
+});
+
+const anOrganizationFiscalCode = "01234567890" as OrganizationFiscalCode;
```
francescopersico: If it is useful and will be used in other tests please create a new shared file.

GetSubscriptionsFeed/__tests__/handler.test.ts
```diff
@@ -0,0 +1,121 @@
...
+    _:
+      "1-2-3-4-5-6-e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+  }
+};
+
+const mockQueryEntityResuls: TableService.QueryEntitiesResult<TableEntry> = {
+  entries: [mockTableEntry]
+};
+
+describe("GetSubscriptionFeedHandler", () => {
```
francescopersico:
```suggestion
describe("GetSubscriptionsFeedHandler", () => {
```

GetSubscriptionsFeed/__tests__/handler.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const result = await getSubscriptionFeedHandler(
+      aUserAuthenticationDeveloper,
+      undefined as any,
+      someUserAttributes,
+      todayDate
+    );
+
+    expect(tableServiceMock.queryEntities).toHaveBeenCalledTimes(3);
+    expect(result.kind).toBe("IResponseSuccessJson");
```
francescopersico: We also need to test the result content is what we expect.

GetSubscriptionsFeed/utils.ts
francescopersico: The PR has been splitted in 2 parts and is already in master. The part that introduce this type is here https://github.com/teamdigitale/io-functions-commons/pull/34/files#diff-6acba14f29206bf08111645a46659d1fR233.

giux78: ok i will import the correct type

GetSubscriptionsFeed/__tests__/handler.test.ts
giux78: shared file in this repo? In case can I put it in the utils folder? Any different conventions?

GetSubscriptionsFeed/__tests__/handler.test.ts
giux78: ok

GetSubscriptionsFeed/__tests__/handler.test.ts
giux78: I don't know the expected behaviour, at the moment I think but I'm not completely sure also an empty array can be a valid result from the API, and in this test is the result content is effectively an empty array. Do you think I can check for that? For more complex result and test also the API should be changed. IMHO.

GetSubscriptionsFeed/__tests__/handler.test.ts
francescopersico: In a __mocks__ directory like https://github.com/teamdigitale/io-functions-admin/blob/master/__mocks__/mocks.ts

GetSubscriptionsFeed/__tests__/handler.test.ts
francescopersico: If you have to explain to another person what GetSubscriptionsFeedHandler function MUST do to return the right result probably you say something like:
1) you have to query the db to get all the PROFILE SUBSCRIPTION
2) you have to query the db to get all the SERVICE SUBSCRIPTION
3) you have to query the db to get all the SERVICE UNSUBSCRIPTION
....

What you have to test is that
1) all these steps get correctly executed
2) simulating (mocking) the result of each step the output is what you expect, for example if you simulate that for a user in the X day there is only a SERVICE UNSUBSCRIPTION the return value must include the hashed fiscalcode of the user in the unsubscriptions array.

PR #26, opened 2019-11-12
[#169703944] Disable email channel if profile email is not validated

francescopersico:


CreateNotificationActivity/handler.ts
```diff
@@ -157,6 +157,9 @@ export const getCreateNotificationActivityHandler = (
   // in their profile
   const isEmailEnabledInProfile = profile.isEmailEnabled;
 
+  // Check if the email in the user profile is validated
+  const isEmailValidatedInProfile = profile.isEmailValidated;
```
gunzip: a question: is this set to true by default when using SPID login ?

francescopersico: Yes on login the user profile is created with
1) is_email_validated = true if logged in with SPID
2) is_email_validated = false if logged in with CIE

PR #28, opened 2019-12-05
Revert "[#169703944] Disable email channel if profile email is not validated"

francescopersico:
Reverts teamdigitale/io-functions-services#26


gunzip:
@francescopersico why reverting ?

PR #42, opened 2020-04-07
[#172163927] add keepalive and timeout to webhook calls

gunzip:
drop superagent and use node-fetch with timeout and keepalive to call the notify webhook

### DIFF
--- a/WebhookNotificationActivity/client.ts
+++ b/WebhookNotificationActivity/client.ts
@@ -1,0 +1,37 @@
+import { HttpsUrl } from "io-functions-commons/dist/generated/definitions/HttpsUrl";
+import {
+  ApiHeaderJson,
+  createFetchRequestForApi,
+  TypeofApiCall
+} from "italia-ts-commons/lib/requests";
+import * as r from "italia-ts-commons/lib/requests";
+import { ProblemJson } from "italia-ts-commons/lib/responses";
+import { Notification } from "../generated/notifications/Notification";
+import { notifyDefaultDecoder } from "../generated/notifications/requestTypes";
+import { SuccessResponse } from "../generated/notifications/SuccessResponse";
+
+export type WebhookNotifyT = r.IPostApiRequestType<
+  { readonly notification?: Notification; readonly webhookEndpoint: HttpsUrl },
+  "Content-Type",
+  never,
+  // tslint:disable-next-line: max-union-size
+  | r.IResponseType<200, SuccessResponse>
+  | r.IResponseType<400, ProblemJson>
+  | r.IResponseType<401, undefined>
+  | r.IResponseType<500, ProblemJson>
+>;
+
+export const getNotifyClient = (
+  fetchApi: typeof fetch
+): TypeofApiCall<WebhookNotifyT> =>
+  createFetchRequestForApi(
+    {
+      body: params => JSON.stringify(params.notification),
+      headers: ApiHeaderJson,
+      method: "post",
+      query: _ => ({}),
+      response_decoder: notifyDefaultDecoder(),
+      url: params => `${params.webhookEndpoint}`
+    } as WebhookNotifyT,
+    { fetchApi }
+  );
--- a/WebhookNotificationActivity/client.ts
+++ b/WebhookNotificationActivity/client.ts
@@ -11,7 +11,7 @@ import { notifyDefaultDecoder } from "../generated/notifications/requestTypes";
 import { SuccessResponse } from "../generated/notifications/SuccessResponse";
 
 export type WebhookNotifyT = r.IPostApiRequestType<
-  { readonly notification?: Notification; readonly webhookEndpoint: HttpsUrl },
+  { readonly notification: Notification; readonly webhookEndpoint: HttpsUrl },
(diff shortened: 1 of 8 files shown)


WebhookNotificationActivity/client.ts
```diff
@@ -0,0 +1,37 @@
...
+  TypeofApiCall
+} from "italia-ts-commons/lib/requests";
+import * as r from "italia-ts-commons/lib/requests";
+import { ProblemJson } from "italia-ts-commons/lib/responses";
+import { Notification } from "../generated/notifications/Notification";
+import { notifyDefaultDecoder } from "../generated/notifications/requestTypes";
+import { SuccessResponse } from "../generated/notifications/SuccessResponse";
+
+export type WebhookNotifyT = r.IPostApiRequestType<
+  { readonly notification?: Notification; readonly webhookEndpoint: HttpsUrl },
```
cloudify, line 14: why is `notification` optional? what's the use case for an empty notification?

WebhookNotificationActivity/index.ts
```diff
@@ -44,14 +53,26 @@ const notificationModel = new NotificationModel(
 // Whether we're in a production environment
 const isProduction = process.env.NODE_ENV === "production";
 
+// Webhook must be an https endpoint
+const abortableFetch = AbortableFetch(agent.getHttpsFetch(process.env));
```
cloudify, line 57: why `getHttpsFetch` needs to have access to all the environment? we should minimize the data required, can we just pass the relevant env vars?

WebhookNotificationActivity/index.ts
```diff
@@ -44,14 +53,26 @@ const notificationModel = new NotificationModel(
...
+// 10 seconds timeout by default
+const fetchWithTimeout = setFetchTimeout(
```
cloudify, line 59: this behavior differs from the one we implemented in `app-backend` where we rely on `agentkeepalive`'s defaults, I would try to keep the behavior consistent across components

WebhookNotificationActivity/index.ts
```diff
@@ -44,14 +53,26 @@ const notificationModel = new NotificationModel(
...
+  (process.env.FETCH_KEEPALIVE_TIMEOUT
```
cloudify, line 60: I'm not sure this makes sense, `FETCH_KEEPALIVE_TIMEOUT` sets the [timeout on the socket](https://github.com/node-modules/agentkeepalive#new-agentoptions) while `setFetchTimeout` sets the timeout on the HTTP(s) request - I believe the former should be much higher than the latter (in fact it defaults to 30s)

WebhookNotificationActivity/client.ts
gunzip: The generated requestType (from io-utils) is:

```ts
// Request type definition
export type NotifyT = r.IPostApiRequestType<
  { readonly notification?: Notification },
  "Content-Type",
  never,
  | r.IResponseType<200, SuccessResponse>
  | r.IResponseType<400, ProblemJson>
  | r.IResponseType<401, undefined>
  | r.IResponseType<500, ProblemJson>
>;
```

probably because this parameter is not required
https://github.com/pagopa/io-backend/blob/master/api_notifications.yaml#L20

WebhookNotificationActivity/index.ts
gunzip: the whole point of this utility method is to abstract away the processing of options from the environment.

if we make it accepts only the required properties we end up reply (cut&paste) all this logic wherever you have to get a new agent: https://github.com/pagopa/io-ts-commons/blob/master/src/agent.ts#L10

which make the whole agent module useless

WebhookNotificationActivity/index.ts
gunzip: what do you mean exactly ? the configure http agent does not terminate (abort) the connection, it only signal a timeout event and keeps the connection open. to close it we must use `setFetchTimeout`

WebhookNotificationActivity/index.ts
gunzip: what do you suggest as request timeout value ?

WebhookNotificationActivity/client.ts
cloudify: it looks like it's [actually required](https://github.com/pagopa/io-backend/blob/master/src/controllers/notificationController.ts#L49), we should fix the specs

WebhookNotificationActivity/index.ts
cloudify: ok I see 👍

WebhookNotificationActivity/index.ts
cloudify: sorry I misunderstood the code

WebhookNotificationActivity/index.ts
cloudify: I would look at the 95th percentile response time of the webhook endpoint and add 20% to that

WebhookNotificationActivity/client.ts
gunzip: fixed in io-backend

PR #47, opened 2020-04-18
[#172332578] singleton cosmosdb client

gunzip:


utils/cosmosdb.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Use a singleton CosmosDB client across functions.
+ */
+import { DocumentClient as DocumentDBClient } from "documentdb";
+
+// tslint:disable-next-line: no-let
+let instance: DocumentDBClient;
+
+export function getDocumentClient(
```
cloudify, line 9: This is actually not a good implementation - it works in our case since the params don't change, but it's misleading and can cause bugs in the future - `instance` will be created with the `cosmosDbUri` and `cosmosDbKey` provided in the first call - these parameters will be ignored on subsequent calls - so say somebody tries to call this function to get the client for multiple cosmosdb URIs, it will not work as expected. 

The problem is that a singleton is supposed to receive the creation params once (on constructor).

You have two options:

1) (simpler, preferred) instantiate a `DocumentDBClient` right away in this file and export it, i.e.

```typescript
const cosmosDbUri = getRequiredStringEnv("COSMOSDB_URI");
const masterKey = getRequiredStringEnv("COSMOSDB_KEY");

export const documentDbClient =  new DocumentDBClient(cosmosDbUri, { masterKey }));
```

2) (little more complex but doesn't add much) make it a singleton class, and do as in (1)

3) (more complex, not needed since we have only one URL) make `getDocumentClient` keep a map of `cosmosDbUri` → `DocumentDBClient` and make it instantiate/return the instance associated to the URL

cloudify:
@gunzip sorry I was late in reviewing you code

utils/cosmosdb.ts
gunzip: you're right I'm fixing it [here](https://github.com/pagopa/io-functions-services/pull/50/files) and for io-functions-app

PR #51, opened 2020-04-23
[#172220283] Use commons agent for MailUP calls

cloudify:
Need to fix the missing `initAppInsights` in the latest `io-functions-commons`


cloudify:
@gunzip it's a draft because of https://github.com/pagopa/io-functions-services/pull/51#issue-408012287

PR #55, opened 2020-05-11
172766902 rate limits info

francescopersico:
Draft until the limits are approved

### DIFF
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -1,6 +1,6 @@
 swagger: '2.0'
 info:
-  version: 1.0.0
+  version: 1.0.1
   title: IO API for Public Administration Services
   contact:
     name: Digital Transformation Team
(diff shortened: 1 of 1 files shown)


openapi/index.yaml
```diff
@@ -34,6 +34,13 @@ info:
     sent to the fake Fiscal Code will be notified to the email address used
     during the registration process on the developer portal.
 
+    # Rate limits
+
+    All API requests are subject to rate limits.
+    Rate limits apply on a subscription key basis.
+    The current rate limit is of `150 requests` every `5 seconds`.
+    If limits are exceeded the API server returns a 429 (Too Many Requests) response.
```
gunzip, line 42: do we return some headers to the caller ? (ie x-retry-after)

francescopersico: Not for now, but we can add both retry-after and remaining-calls count as headers.

gunzip: ok (so we will add those headers here when ready 👍 )

gunzip: do we return the headers now? :)

gunzip: ping :)

PR #62, opened 2020-07-27
[#171554236] Update io-functions-commons to v12

balanza:
Update `io-functions-commons` to v12 that ships CosmosDB SDK update.

### Changes
* updated `io-functions-commons` dependency
* changed the way we build Cosmos client from configuration, and adapted model instances accordingly
* added `.run()` to model methods that used to have a return type of `Promise<Either<L,R>` and now is`TaskEither<L,R>`.
* replaced removed `findOneProfileByFiscalCode` method with `findLastVersionByModelId`
* ~applied manual casting from `string` to `NonEmptyString` for document ids (this might be obsolete when we merge https://github.com/pagopa/io-functions-commons/pull/81)~
* replaced manual cast to `NonNegativeNumber` with `NonNegativeInteger` in versioned model definitions 
* fixed test mocks according to new methods signatures


CreateMessage/index.ts
```diff
@@ -30,26 +29,26 @@ secureExpressApp(app);
...
-);
-const servicesCollectionUrl = documentDbUtils.getCollectionUri(
-  documentDbDatabaseUrl,
-  SERVICE_COLLECTION_NAME
-);
+const cosmosdbClient = new CosmosClient({
+  endpoint: cosmosDbUri,
+  key: cosmosDbKey
+});
+
```
AleDore, line 41: Maybe, we can extract and refactor this code into a util file

CreateNotificationActivity/index.ts
```diff
@@ -42,17 +40,17 @@ const emailNotificationServiceBlackList = parseCommaSeparatedListOf(ServiceId)(
...
 // Setup DocumentDB
+const cosmosDbUri = getRequiredStringEnv("COSMOSDB_URI");
 const cosmosDbName = getRequiredStringEnv("COSMOSDB_NAME");
+const cosmosDbKey = getRequiredStringEnv("COSMOSDB_KEY");
 
-const documentDbDatabaseUrl = documentDbUtils.getDatabaseUri(cosmosDbName);
+const cosmosdbClient = new CosmosClient({
+  endpoint: cosmosDbUri,
+  key: cosmosDbKey
+});
```
AleDore, line 50: ditto

CreateMessage/index.ts
```diff
@@ -30,26 +29,26 @@ secureExpressApp(app);
...
 const messageModel = new MessageModel(
-  documentClient,
-  messagesCollectionUrl,
+  cosmosdbClient.database(cosmosDbName).container(MESSAGE_COLLECTION_NAME),
   messageContainerName
 );
 
-const serviceModel = new ServiceModel(documentClient, servicesCollectionUrl);
+const serviceModel = new ServiceModel(
+  cosmosdbClient.database(cosmosDbName).container(SERVICE_COLLECTION_NAME)
```
AleDore, line 50: Also `database` can be moved into a util file.

CreateMessage/index.ts
gunzip: yes, there should be only one singleton client shared between functions

GetMessage/__tests__/handler.test.ts
```diff
@@ -117,8 +117,13 @@ const aNewMessageWithoutContent: NewMessageWithoutContent = {
...
 const aRetrievedMessageWithoutContent: RetrievedMessageWithoutContent = {
   ...aNewMessageWithoutContent,
+  /* 
+  FIXME: these fields are shipped with RetrievedNotificationStatus objects
+  However, RetrievedNotificationStatus type doesn't include such fields in its definition
+  hence, the compiler consider such fields an error
   _self: "xyz",
-  _ts: 1,
+  _ts: 1, 
+  */
```
balanza, line 126: @cloudify here is a dilemma: 
attributes such as `_self` are present in retrieved objects, but we actually never use them. The error above is due to this configuration: https://github.com/pagopa/io-functions-commons/blob/master/src/utils/cosmosdb_model.ts#L124

Should we refactor `CosmosdbModel` to extend `ResourceT`? I think such metadata is useless at user level, but still we have an incorrect structural mapping

balanza:
@gunzip tested with io-mock and it seems fine. We can merge imo.

StoreMessageContentActivity/index.ts
```diff
@@ -9,30 +10,15 @@ import {
...
   ProfileModel
 } from "io-functions-commons/dist/src/models/profile";
-import * as documentDbUtils from "io-functions-commons/dist/src/utils/documentdb";
 import { getRequiredStringEnv } from "io-functions-commons/dist/src/utils/env";
-import { documentClient } from "../utils/cosmosdb";
 import { getStoreMessageContentActivityHandler } from "./handler";
 
-// Setup DocumentDB
-const cosmosDbName = getRequiredStringEnv("COSMOSDB_NAME");
+const x = cosmosdbInstance.container(PROFILE_COLLECTION_NAME);
```
francescopersico, line 16: x?

PR #80, opened 2020-10-08
[#175119553] Centralize environment variables in a single config module

infantesimone:
This PR introduces the utils/config module which aims to be the unique interface for application configuration.

Every place in which the application used to directly access env variable has been refactored to use this module instead.


balanza (approved):
lgtm

AleDore (commented):
lgtm

PR #81, opened 2020-10-08
[#175185887] Upgrade Italia-utils to allow Api auth without bearer

AleDore:
This PR is intended to avoid generated api clients to use `Bearer` token authentication while calling io-functions-admin APIs


gunzip (changes requested):
you must upgrade io-fn-commons to 15.x as well (otherwise it does not build)

PR #82, opened 2020-10-23
[#175205379] Added mailing component from io-functions-commons

infantesimone:
This PR introduces the mailing component developed [here](https://github.com/pagopa/io-functions-commons/pull/112).


EmailNotificationActivity/handler.ts
```diff
@@ -146,7 +148,8 @@ export const getEmailNotificationActivityHandler = (
 
   // trigger email delivery
   // see https://nodemailer.com/message/
-  const sendResult = await sendMail(lMailerTransporter, {
+
+  sendMail(lMailerTransporter, {
```
gunzip, line 152: I guess this returns a promise, shouldn't we await here?

balanza: @gunzip no, `sendMail` changed. Now it returns a `TaskEither`

EmailNotificationActivity/handler.ts
```diff
@@ -160,16 +163,12 @@ export const getEmailNotificationActivityHandler = (
...
-    throw new Error(`Error while sending email: ${error.message}`);
-  }
-
-  context.log.verbose(`${logPrefix}|RESULT=SUCCESS`);
+  })
+    .map(() => context.log.verbose(`${logPrefix}|RESULT=SUCCESS`))
+    .mapLeft(error => {
+      context.log.error(`${logPrefix}|ERROR=${error.message}`);
+      throw new Error(`Error while sending email: ${error.message}`);
+    });
```
balanza, line 171: Aren't you missing a `.run()`?

balanza: (anyway, when you have adjacent `map` and `mapLeft` you can use `bimap`

balanza (changes requested):

EmailNotificationActivity/handler.ts
gunzip: then you need to call `run()` + `await` otherwise this would be a no-action (should be caught by some unit test)

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,146 @@
...
+      message: aMessage,
+      notificationId: aNotificationId,
+      senderMetadata: aSenderMetadata
+    }
+  };
+
+  it("should respond with 'SUCCESS' if the mail is sent", async () => {
+    const lMailerTransporterMock = ({
+      sendMail: jest.fn((_, f) => f(null, taskEither.of("SUCCESS")))
+    } as unknown) as MailerTransporter;
```
balanza, line 131: By mocking the transport's `sendMail` you are _cheating_, as you are using the `mailer`'s inner implementation (something you assume to work). Try mocking `mailer`'s `sendMail` method instead. 
At this level, we only need to test `mailer`'s `sendMail` boundaries (that is: all its possible results).

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,146 @@
...
+  const input: EmailNotificationActivityInput = {
+    notificationEvent: {
+      content: aMessageContent,
+      message: aMessage,
+      notificationId: aNotificationId,
+      senderMetadata: aSenderMetadata
+    }
+  };
+
+  it("should respond with 'SUCCESS' if the mail is sent", async () => {
```
balanza, line 128: Good! Also, can we test error paths as well?

EmailNotificationActivity/index.ts
```diff
@@ -32,6 +31,7 @@ import { NonEmptyString } from "italia-ts-commons/lib/strings";
 import { Millisecond } from "italia-ts-commons/lib/units";
 import nodemailerSendgrid = require("nodemailer-sendgrid");
 
+import { MailUpTransport } from "io-functions-commons/dist/src/mailer/transports";
```
balanza, line 34: You actually don't want to manually create a transport. Instead, we just want to call the `mailer`'s factory metodo for that (see [here](https://github.com/pagopa/io-functions-commons/pull/112#pullrequest-500583371-permalink:~:text=The%20folder%20has%20a%20index.ts%20entrypoint,proxied%20by%20the%20index%20module%20itself.)).

```suggestion
import { getMailerTransporter } from "io-functions-commons/dist/src/mailer";
```

balanza (changes requested):

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,146 @@
...
+    const GetEmailNotificationActivityHandler = getEmailNotificationActivityHandler(
+      lMailerTransporterMock,
+      notificationModelMock,
+      defaultNotificationParams
+    );
+
+    const result = (await GetEmailNotificationActivityHandler(
+      mockContext,
+      input
+    )) as EmailNotificationActivityResult;
```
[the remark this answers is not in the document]
infantesimone: something like that?

 ```
 EmailNotificationActivityResult.decode(result).map(result =>
      expect(result.kind).toBe("SUCCESS")
    );
```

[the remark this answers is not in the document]
Comment (balanza) [reply EmailNotificationActivity/__tests__/handler.test.ts:142]: there's should not be the need for a decode. Actually the "problem" is `result` to be inferred as `unknown`. Which is correct, as [the function itself is marked](https://github.com/pagopa/io-functions-services/blob/9649d738a59d773204343dd8301306ba49759a73/EmailNotificationActivity/handler.ts#L57) as `Promise<unknown>`. 
I think it's better to have such function being defined as `Promise<EmailNotificationActivityResult>`.

Why am I asking this?
Types are our safe net. If you need to force the cast, it should sound as an alert to you that something isn't  composed as you expected. So first thing to do is trying to live without them. If you really can't - and it happens sometimes - we should motivate that by adding a comment (like [here](https://github.com/pagopa/io-functions-commons/blob/c46d77a5e5e8175d05d57dfc98cbac11c0661f7d/src/utils/cosmosdb_model_versioned.ts#L200) for example).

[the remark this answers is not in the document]
Comment (infantesimone) [reply EmailNotificationActivity/__tests__/handler.test.ts:142]: ook 💪

File: EmailNotificationActivity/__tests__/handler.test.ts (hunk 3 above)
Comment (infantesimone) [reply EmailNotificationActivity/__tests__/handler.test.ts:131]: but you need a mock (`lMailerTransporterMock`) for getEmailNotificationActivityHandler:

   ```
const emailNotificationActivityHandler = getEmailNotificationActivityHandler(
      lMailerTransporterMock,
      notificationModelMock,
      defaultNotificationParams
    );

```

balanza: you can mock it as well, just passing anything that comply with its type definition. However, you are not supposed to play with it's implementation to test `mailer.sendMail`

EmailNotificationActivity/__tests__/handler.test.ts
infantesimone: When sending mail fails, this is executed:
```
 context.log.error(`${logPrefix}|ERROR=${error.message}`);
 throw new Error(`Error while sending email: ${error.message}`);
```
is there a particular reason (maybe due to Azure Activity)? because coding the tests I thought it could be simplified/uniformed like this:

```
return EmailNotificationActivityResult.encode({
    kind: "FAILURE",
    result: "OK"
  });
```

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,165 @@
...
+import { CreatedMessageEventSenderMetadata } from "io-functions-commons/dist/src/models/created_message_sender_metadata";
+import {
+  NewNotification,
+  NotificationAddressSourceEnum,
+  NotificationModel,
+  RetrievedNotification
+} from "io-functions-commons/dist/src/models/notification";
+
+describe("getEmailNotificationActivityHandler", () => {
+  const mockContext = {
```
balanza, line 36: Add this just to be sure tests are isolated (mocked values and implementations persist for all tests in a file)
```suggestion
  beforeEach(() => jest.clearAllMocks());
  const mockContext = {
```

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,165 @@
...
+import * as mail from "io-functions-commons/dist/src/mailer";
+import { CreatedMessageEventSenderMetadata } from "io-functions-commons/dist/src/models/created_message_sender_metadata";
+import {
+  NewNotification,
+  NotificationAddressSourceEnum,
+  NotificationModel,
+  RetrievedNotification
+} from "io-functions-commons/dist/src/models/notification";
+
+describe("getEmailNotificationActivityHandler", () => {
```
balanza, line 35: just a note on style:

try to put mocked values outside the `describe` instruction (*). This way the test case is concise and more readable


(*) best: in a `__mocks__/mocks.ts` file, but it's ok for now to leave them here.

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      message: aMessage,
+      notificationId: aNotificationId,
+      senderMetadata: aSenderMetadata
+    }
+  };
+
+  const lMailerTransporterMock = ({} as unknown) as mail.MailerTransporter;
+
+  it("should respond with 'SUCCESS' if the mail is sent", async () => {
+    jest.spyOn(mail, "sendMail").mockReturnValue(taskEither.of("SUCCESS"));
```
balanza, line 130: Good one, that's exactly the point of testing `sendMail` as a black box. 
Just a note: when mocking in jest, try to use `mock{something}Once` (note the _Once_) when possible; in this case, `mockReturnValueOnce`. 
```suggestion
    jest.spyOn(mail, "sendMail").mockReturnValueOnce(taskEither.of("SUCCESS"));
```

The reason is that, when you override a function's implementation, this applies for every successive calls to such functions, which means that the test order matter (and we want our tests to be independent). 

(Please note there is a bit of cheating here: we assume `sendMail` is called just once 😄 . But it's ~fine~ less ugly because it is an assumption on the very code we are testing, not on a third-party module)

EmailNotificationActivity/__tests__/handler.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    expect(result.kind).toBe("SUCCESS");
+  });
+
+  it("should respond with 'ERROR' if the mail is not sent", async () => {
+    const errorMessage: string = "Test Error";
+
+    jest
+      .spyOn(mail, "sendMail")
+      .mockReturnValue(fromLeft(new Error(errorMessage)));
```
balanza, line 151: same as above

EmailNotificationActivity/handler.ts
```diff
@@ -52,7 +54,10 @@ export const getEmailNotificationActivityHandler = (
   lMailerTransporter: NodeMailer.Transporter,
   lNotificationModel: NotificationModel,
   notificationDefaultParams: INotificationDefaults
-) => async (context: Context, input: unknown): Promise<unknown> => {
+) => async (
+  context: Context,
+  input: unknown
+): Promise<EmailNotificationActivityResult> => {
```
balanza, line 60: Big 🆙

balanza (changes requested):

balanza (approved):
LGTM, if you rebase with master we can merge

PR #83, opened 2020-10-26
[#175395536] Add Organization logo's upload

AleDore:
This PR adds a new endpoint for `UploadOrganizationLogo`operation. See [io-functions-admin PR](https://github.com/pagopa/io-functions-admin/pull/96) for further informations


openapi/index.yaml
```diff
@@ -541,6 +541,42 @@ paths:
           description: Cannot regenerate service key.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/organizations/{organization_fiscal_code}/logo":
+    parameters:
+      - name: organization_fiscal_code
+        in: path
+        type: string
```
balanza, line 548: Do we have a better definition than `string`? Can we use [`OrganizationFiscalCode`](https://github.com/pagopa/io-functions-commons/blob/d615113237d2d65a55a9b6e1f81b762f12f9c108/openapi/definitions.yaml#L287)

openapi/index.yaml
```diff
@@ -541,6 +541,42 @@ paths:
...
+      parameters:
+        - name: body
+          in: body
+          required: true
+          schema:
+            $ref: "#/definitions/Logo"
+          description: A base64 string representation of the organization logo PNG image.
+      responses:
+        "200":
+          description: Logo uploaded.
```
balanza, line 567: Do we prefer/can return 201 along with the reference to the just uploaded image?

openapi/index.yaml
```diff
@@ -541,6 +541,42 @@ paths:
...
+          schema:
+            $ref: "#/definitions/ProblemJson"
+        "401":
+          description: Unauthorized.
+        "403":
+          description: Forbidden.
+        "500":
+          description: The service logo cannot be uploaded.
+          schema:
+            $ref: "#/definitions/ProblemJson"
```
balanza, line 579: Maybe we want to add 404 in case no `Organization` is found with such fiscal code?

package.json
```diff
@@ -17,7 +17,7 @@
     "lint": "tslint -p .",
     "generate:definitions": "rimraf ./generated/definitions && shx mkdir -p ./generated/definitions && gen-api-models --api-spec ./openapi/index.yaml --no-strict --out-dir ./generated/definitions",
     "generate:api-notifications": "rimraf ./generated/notifications && shx mkdir -p ./generated/notifications && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-backend/master/api_notifications.yaml --out-dir ./generated/notifications --response-decoders --request-types",
-    "generate:api-admin": "rimraf generated/api-admin && shx mkdir -p generated/api-admin && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-functions-admin/master/openapi/index.yaml --no-strict --out-dir generated/api-admin --request-types --response-decoders --client",
+    "generate:api-admin": "rimraf generated/api-admin && shx mkdir -p generated/api-admin && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-functions-admin/175395536_upload_organization_logo/openapi/index.yaml --no-strict --out-dir generated/api-admin --request-types --response-decoders --client",
```
balanza, line 20: This should not have been committed

UploadOrganizationLogo/handler.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Logo } from "../generated/api-admin/Logo";
+import { withApiRequestWrapper } from "../utils/api";
+import { getLogger, ILogger } from "../utils/logging";
+import { ErrorResponses, IResponseErrorUnauthorized } from "../utils/responses";
+
+type ResponseTypes =
+  | IResponseSuccessJson<undefined>
+  | IResponseErrorUnauthorized
+  | IResponseErrorForbiddenNotAuthorized
+  | IResponseErrorNotFound
```
balanza, line 54: We must map it in the spec as well

UploadOrganizationLogo/handler.ts
```diff
@@ -0,0 +1,132 @@
...
+  | IResponseErrorTooManyRequests
```
balanza, line 55: We must map it in the specs as well

UploadOrganizationLogo/handler.ts
```diff
@@ -0,0 +1,132 @@
...
+  organizationFiscalCode: OrganizationFiscalCode,
+  logoPayload: Logo
+) => Promise<ResponseTypes>;
+
+const uploadOrganizationLogoTask = (
+  logger: ILogger,
+  apiClient: APIClient,
+  organizationFiscalCode: OrganizationFiscalCode,
+  logo: Logo
+): TaskEither<ErrorResponses, IResponseSuccessJson<undefined>> =>
```
balanza, line 80: SInce we are returning `undefined`, is that a hint that we might return something else (like 202 o0r 204)?

UploadOrganizationLogo/__tests__/handler.test.ts
```diff
@@ -0,0 +1,160 @@
...
+  subscriptionId: aServiceId,
+  userId: "u123" as NonEmptyString
+};
+
+const aLogoPayload: Logo = {
+  logo: "base64-logo-img" as NonEmptyString
+};
+
+describe("UploadOrganizationLogo", () => {
+  it("should respond with 201 if logo upload was successfull", async () => {
```
balanza, line 83: Why 201? I think is 200, according to the implementation

UploadOrganizationLogo/__tests__/handler.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    const result = await uploadOrganizationLogoHandler(
+      mockContext,
+      aUserAuthenticationDeveloper,
+      undefined as any, // not used
+      someUserAttributes,
+      anOrganizationFiscalCode,
+      aLogoPayload
+    );
+
+    expect(apiClientMock.uploadOrganizationLogo).toHaveBeenCalledTimes(1);
```
balanza, line 102: Do we need this test? This function is mocked, so we expect the test to just fail if it's not called

UploadOrganizationLogo/__tests__/handler.test.ts
```diff
@@ -0,0 +1,160 @@
...
```
balanza, line 128: ditto

UploadOrganizationLogo/__tests__/handler.test.ts
```diff
@@ -0,0 +1,160 @@
...
```
balanza, line 156: ditto

balanza (changes requested):

balanza (changes requested):
I made a lot of comments and they might appear unrelated and conflicting. Here I try to sum up:
* Implementation and specification are not aligned
* We might want to model the success case with something different than 200
* Can we use `OrganizationFiscaleCode` definition?
* We should avoid testing implementation (that is, _has this been called_) and try to test input/output instead

openapi/index.yaml
AleDore: Nope, there's another [story](https://www.pivotaltracker.com/story/show/175436161) to achieve this. At the moment we don't have a check mechanism cause Organization Model is missing.

UploadOrganizationLogo/__tests__/handler.test.ts
AleDore: Yes, this must be corrected 👍

package.json
AleDore: Yep, It's a Draft for this reason :)

openapi/index.yaml
AleDore: The io-functions-admin endpoint, that is called inside this endpoint on io-functions-services' side, returns 201 with the link to the uploaded image. I don't remember well, but in this case we are avoiding to expose infos about storage and CDN links to external subjects. cc: @gunzip

openapi/index.yaml
balanza: The point is that the handler can actually return 404

AleDore:
Can we merge this? @gunzip @balanza

UploadOrganizationLogo/handler.ts
```diff
@@ -0,0 +1,140 @@
...
+    () =>
+      apiClient.uploadOrganizationLogo({
+        body: logo,
+        organization_fiscal_code: organizationFiscalCode
+      }),
+    201
+  ).map(_ => ResponseSuccessAccepted());
+
+/**
+ * Handles requests for upload a service logo by a service ID and a base64 logo' s string.
```
gunzip, line 94:
```suggestion
 * Handles requests for upload an organization logo.
```

openapi/index.yaml
```diff
@@ -541,6 +541,40 @@ paths:
...
             $ref: "#/definitions/ProblemJson"
+  "/organizations/{organization_fiscal_code}/logo":
+    parameters:
+      - $ref: "#/parameters/OrganizationFiscalCode"
+    put:
+      summary: Upload organization logo.
+      description: |
+        Upsert a logo for an Organization.
+        This API operation needs a valid API key
+        otherwise 403 forbidden will be returned to the caller.
```
gunzip, line 552: this is implied by the 403 status below

openapi/index.yaml
```diff
@@ -684,6 +718,17 @@ parameters:
...
+    name: organization_fiscal_code
+    in: path
+    type: string
+    description: Organization fiscal code.
+    format: OrganizationFiscalCode
+    minLength: 11
+    maxLength: 11
+    pattern: "^[0-9]{11}$"
+    required: true
+    x-example: "12345678901"
```
gunzip, line 731: you should use this format:

https://github.com/pagopa/io-functions-commons/blob/master/openapi/definitions.yaml#L290

gunzip (changes requested):

gunzip (approved):

PR #85, opened 2020-10-28
[#175160851] Block create/update of a visible service without the metadata,

infantesimone:
This PR introduces the new [Service](https://raw.githubusercontent.com/pagopa/io-functions-commons/v17.0.0/openapi/definitions.yaml#/Service) definition thats blocks update/create of a visible service without metadata.


balanza:
This is an important rule that ensures that our data are consistent, and we should try to have it to work ad the lowest level possible. 

Let's tackle this problem with a different approach, let's try to encode such rule right inside the model definition.

When you have a flag, this usually splits your model `MyModel` into `MyModelWithTrueFlag` and `MyModelWithTFalseFlag`, so your model becomes `MyModel = MyModelWithTrueFlag | MyModelWithTFalseFlag`. 
As for our need, we can define the `Service` model as following:
```ts
const VisibleService = t.interface({ isVisible: t.literal(true), service_metadata: ServiceMetadata });
const InvisibleService = t.intersection([
    t.interface({ isVisible: t.literal(false) }),
    t.partial({ service_metadata: ServiceMetadata  })
]);
const Service = t.union([VisibleService, VisibleService]);
```
_(Please note I'm not including every field.)_

Now you don't need to add extra validation steps: every time you try to decode a Service from an arbitrary input payload, it checks that `if the service is visible, it must have metadata, otherwise they can be either present or not`. 
This technique has two major benefits:
1. rules are applied everywhere in our ecosystem (including other apps)
2. the difference between a visible and an invisible service are now explicit, you just need to look the model

Do you think it could be a good way to approach this problem?

AleDore:
> This is an important rule that ensures that our data are consistent, and we should try to have it to work at the lowest possible level.
> 
> Let's tackle this problem with a different approach, let's try to encode such rule right inside the model definition.
> 
> When you have a flag, this usually splits your model `MyModel` into `MyModelWithTrueFlag` and `MyModelWithTFalseFlag`, so your model becomes `MyModel = MyModelWithTrueFlag | MyModelWithTFalseFlag`.
> As for our need, we can define the `Service` model as following:
> 
> ```ts
> const VisibleService = t.interface({ isVisible: t.literal(true), service_metadata: ServiceMetadata });
> const InvisibleService = t.intersection([
>     t.interface({ isVisible: t.literal(false) }),
>     t.partial({ service_metadata: ServiceMetadata  })
> ]);
> const Service = t.union([VisibleService, VisibleService]);
> ```
> 
> _(Please note I'm not including every field.)_
> 
> Now you don't need to add extra validation steps: every time you try to decode a Service from an arbitrary input payload, it checks that `if the service is visible, it must have metadata, otherwise they can be either present or not`.
> This technique has two major benefits:
> 
> 1. rules are applied everywhere in our ecosystem (including other apps)
> 2. the difference between a visible and an invisible service are now explicit, you just need to look the model
> 
> Do you think it could be a good way to approach this problem?

It's a good point that could be used in this PR, to avoid ternary operators. In order to apply this rule in other apps, I see 2 possibilities:

- Define Models with flag logic into io-functions-commons and let them to be used by others functions that includes commons as a lib
- Try to replicate this logic using Italia-utils while generating models from open api's specs (I don't know if it's possible)

utils/subscription.ts
```diff
@@ -11,3 +15,25 @@ export const serviceOwnerCheckTask = (
     (svcId: NonEmptyString) => svcId === ownerSubscriptionId,
     _ => ResponseErrorForbiddenNotAuthorized
   )(serviceId);
+
+/**
+ * This check verifies that:
+ *  - isVisible === false || (isVisible === true && svcMetadata !== undefined)
```
AleDore, line 21: Try to avoid write code into comments. Let's explain the function's purpose using a discursive approach.

balanza:
> * Define Models with flag logic into io-functions-commons and let them to be used by others functions that includes commons as a lib

To do so, you just have to edit this file: https://github.com/pagopa/io-functions-commons/blob/d615113237d2d65a55a9b6e1f81b762f12f9c108/src/models/service.ts#L105

> * Try to replicate this logic using Italia-utils while generating models from open api's specs (I don't know if it's possible)

Why may we need it? Btw it is possible

CreateService/handler.ts
```diff
@@ -56,14 +57,16 @@ import { ServiceWithSubscriptionKeys } from "../generated/definitions/ServiceWit
...
 
 type ResponseTypes =
   | IResponseSuccessJson<ServiceWithSubscriptionKeys>
   | IResponseErrorUnauthorized
   | IResponseErrorForbiddenNotAuthorized
   | IResponseErrorNotFound
   | IResponseErrorTooManyRequests
-  | IResponseErrorInternal;
+  | IResponseErrorInternal
+  | IResponseErrorValidation;
```
AleDore, line 69: Let's take care that `400` is mapped into openapi spec file.

RegenerateServiceKey/handler.ts
```diff
@@ -55,7 +56,8 @@ type ResponseTypes =
   | IResponseErrorForbiddenNotAuthorized
   | IResponseErrorNotFound
   | IResponseErrorTooManyRequests
-  | IResponseErrorInternal;
+  | IResponseErrorInternal
+  | IResponseErrorValidation;
```
AleDore, line 60: Ditto. Btw Do we need this in `RegenerateServiceKey` ?

UpdateService/handler.ts
```diff
@@ -49,15 +50,19 @@ import { ServiceWithSubscriptionKeys } from "../generated/definitions/ServiceWit
...
 
 type ResponseTypes =
   | IResponseSuccessJson<ServiceWithSubscriptionKeys>
   | IResponseErrorUnauthorized
   | IResponseErrorForbiddenNotAuthorized
   | IResponseErrorNotFound
   | IResponseErrorTooManyRequests
-  | IResponseErrorInternal;
+  | IResponseErrorInternal
+  | IResponseErrorValidation;
```
AleDore, line 65: ditto

UpdateService/handler.ts
```diff
@@ -158,33 +163,38 @@ export function UpdateServiceHandler(
...
-                getLogger(_, logPrefix, "UpdateService"),
-                apiClient,
-                servicePayload,
-                serviceId,
-                retrievedService,
-                userInfo.token_name
-              )
+          serviceVisibleMetadataCheckTask(
+            servicePayload.service_metadata,
+            servicePayload.is_visible
```
AleDore, line 168: If possible, try to move this check after `getUserTask`.

UploadServiceLogo/handler.ts
```diff
@@ -54,7 +55,8 @@ type ResponseTypes =
   | IResponseErrorForbiddenNotAuthorized
   | IResponseErrorNotFound
   | IResponseErrorTooManyRequests
-  | IResponseErrorInternal;
+  | IResponseErrorInternal
+  | IResponseErrorValidation;
```
AleDore, line 59: ditto

AleDore:
> > * Define Models with flag logic into io-functions-commons and let them to be used by others functions that includes commons as a lib
> 
> To do so, you just have to edit this file: https://github.com/pagopa/io-functions-commons/blob/d615113237d2d65a55a9b6e1f81b762f12f9c108/src/models/service.ts#L105
> 
> > * Try to replicate this logic using Italia-utils while generating models from open api's specs (I don't know if it's possible)
> 
> Why may we need it? Btw it is possible

If it is possible we can move this check also on the "client" side. Btw it's not mandatory.

infantesimone:
Based on the above, I think defining `VisibleService` and `InvisibleService` in  `io-functions-commons` is a good solution, 
in order to have (possibly) a double check also on the client side.

utils/subscription.ts
```diff
@@ -11,3 +15,25 @@ export const serviceOwnerCheckTask = (
...
+ * @param isVisible
+ * @param serviceMetadata
+ * @returns serviceMetadata if the check is successful; otherwise ResponseErrorValidation.
+ *
+ */
+export const serviceVisibleMetadataCheckTask = (
```
gunzip, line 27: 1) function that return booleans and variables which are booleans should be prefix with `is` or `has`
(es. `isValidServiceTask`)

2) not required for this PR, but a safer way to make these is to build a io-ts type, somenthing like:

```ts
const ValidServicePayload = t.intersection([ ServicePayload, 
 t.union([
     t.interface({ is_visible: t.literal(false) }) 
     t.interface({ is_visible: t.literal(true), services_metadata: ServiceMetadata })
  ]) 
]);
```

then in your code this task become:

```ts
fromEither(ValidServicePayload.decode(...))
   .chain(updateService)
```

and updateService / createService take `ValidServicePayload` as input.

In this way you're _forced_ to call `ValidServicePayload.decode` before any update or create operation
(while actually they may work with invalid service payloads).

RegenerateServiceKey/handler.ts
infantesimone: `ResponseErrorValidation` is added to `ErrorResponses` of _utilis/responces.ts_,
 so it is added `IResponseErrorValidation` here too, to mach:

- `  .fold<ResponseTypes>(identity, identity)`

in handlers, but it is not correct because the other handlers may not return 400, thanks 💪

infantesimone:
if you agree, for this PR I define `ValidServicePayload` as suggested by @gunzip , and we could evaluate in another PR/story whether to move this definition to io-function-common.

balanza:
> If you agree, for this PR I define `ValidServicePayload` as suggested by @gunzip , and we could evaluate in another PR/story whether to move this definition to io-function-common.

The point is: suppose you have such definition in `io-function-common`, would you still need this PR?

infantesimone:
> > If you agree, for this PR I define `ValidServicePayload` as suggested by @gunzip , and we could evaluate in another PR/story whether to move this definition to io-function-common.
> 
> The point is: suppose you have such definition in `io-function-common`, would you still need this PR?

I think so, because in the case of UpdateService, the service retrieved from the DB is combined with the payload and it is on this combination that the decode / check must be done.

infantesimone:
> > If you agree, for this PR I define `ValidServicePayload` as suggested by @gunzip , and we could evaluate in another PR/story whether to move this definition to io-function-common.
> 
> The point is: suppose you have such definition in `io-function-common`, would you still need this PR?

I think we will need a PR where we will use this new definition.

gunzip:
any update on this one?

gunzip:
ping :-) any update?

gunzip:
@infantesimone it looks like this needs rebasing. is this ready to be deployed then?

infantesimone:
Is to be deployed after [this](https://github.com/pagopa/io-functions-admin/pull/109)

gunzip:
it looks like https://github.com/pagopa/io-functions-admin/pull/109 it's merged :-)

balanza:
@infantesimone can we resolve conflicts? Also, can we merge after?

infantesimone:
rebase done

gunzip:
it looks like we still have conflicts

infantesimone:
Now it should be ok

gunzip (approved):

PR #87, opened 2020-11-04
[#175573274] Allow AI tracking on service's visibility

AleDore:
This PR introduces Application Insights tracking on Create/Update Service APIs with a specific focus to Service visibility, due to security's monitoring purpose.


UpdateService/handler.ts
```diff
@@ -148,9 +150,22 @@ const updateServiceTask = (
...
 ): IUpdateServiceHandler {
   return (_, apiAuth, ___, userAttributes, serviceId, servicePayload) => {
+    const trackUpdateServiceVisibility = (
+      isVisible: boolean,
+      svcId: ServiceId
+    ): void =>
+      telemetryClient.trackEvent({
+        name: "api.services.update",
+        properties: {
+          isVisible: isVisible ? "false" : String(isVisible),
```
balanza, line 164: I'm sure there's meaning on that, but I cannot really get it

UpdateService/handler.ts
```diff
@@ -173,8 +188,11 @@ export function UpdateServiceHandler(
                 userInfo.token_name
               )
             )
-            .chain(service =>
-              getSubscriptionKeysTask(
+            .chain(service => {
```
balanza, line 191: Can we flat this chain-pipeline? In theory, the following
```ts
.chain(
  foo1.chain(
    foo2.chain(
       foo3
    )
  )
)
```

It's equal to 
```ts
.chain( foo1 )
.chain( foo2 )
.chain( foo3 )

File: UpdateService/handler.ts (hunk 1 above)
Comment (AleDore) [reply UpdateService/handler.ts:164]: Nope you're right 🤣

File: UpdateService/handler.ts [hunk 3]
```diff
@@ -161,7 +161,7 @@ export function UpdateServiceHandler(
       telemetryClient.trackEvent({
         name: "api.services.update",
         properties: {
-          isVisible: isVisible ? "false" : String(isVisible),
+          isVisible: isVisible ? String(isVisible) : "false",
```
balanza, line 164: isn't this enough
```suggestion
          isVisible: String(isVisible)
```
or even

```suggestion
          isVisible
```

UpdateService/handler.ts
AleDore: In order to flat chain-pipeline we have to build a struct that contains a subset of right values for each chain, due to usage of this values in the last chain ex:
```ts
.chain(() =>
  foo1.chain(bar1 =>
    foo2.chain(bar2 =>
       foo3(bar1, bar2)
    )
  )
)
```

UpdateService/handler.ts
AleDore: Fixed in [821fa9e](https://github.com/pagopa/io-functions-services/pull/87/commits/821fa9e7bdef60a9b32e956dac12af7d1eacd21a)

UpdateService/handler.ts
```diff
@@ -148,9 +150,22 @@ const updateServiceTask = (
  * Handles requests for updating a service by given serviceId and a Service Payload.
  */
 export function UpdateServiceHandler(
+  telemetryClient: ReturnType<typeof initAppInsights>,
   apiClient: APIClient
 ): IUpdateServiceHandler {
   return (_, apiAuth, ___, userAttributes, serviceId, servicePayload) => {
+    const trackUpdateServiceVisibility = (
```
gunzip, line 157: there's no need to wrap `telemetryClient.trackEvent` into a closure, just use it inline

gunzip (changes requested):

UpdateService/handler.ts
AleDore: Fixed in [6aae3a7](https://github.com/pagopa/io-functions-services/pull/87/commits/6aae3a75463140ef02f6ed99c95de6ba45e9d011)

gunzip (approved):

balanza (approved):
lgtm

PR #89, opened 2020-11-13
[#175119536] Add health check

infantesimone:
This PR introduces the health check in info endpoint,
that check for wrong configuration and unreachable resources (cosmosdb, storage, urls)


balanza (approved):
LGTM

PR #96, opened 2020-11-30
[#175476527] Release stage and pipeline refactor

balanza:
This PR introduces a `Release` stage in the deploy pipeline as defined in https://github.com/pagopa/io-functions-admin/pull/107.

Moreover, a substantial refactor to pipelines has been introduced in the process. Here's key points:
* We have different pipelines for PR-checks and Deploy. This is because those scenarios have different needs in terms of permissions (deploy need to push-back on master branch) and execution strategy (deploys must be sequential, while checks must be done in parallel).
* Everything related to pipelines has been moved to a `.devops` folder
* Deploys are now considered `deployToStagingSlotAndSwap` only, and the deploy stage is way more linear. This led us to remove the local `deploy-steps.yaml` template and instead just describe its steps in the job definition.
* As each pipeline fits more accurately the scenario in which is triggered, there's no longer need for many parameters.
* `circleci` legacy references has been removed


.devops/deploy-pipelines.yml
```diff
@@ -0,0 +1,152 @@
...
+      - patch
+    default: minor
+
+# This pipeline can be manually run or is automatically triggered whenever one 
+# of the following conditions is true:
+# - a push is made to any branch in the repository (not only 'master')
+# - a pull request is created
+# - a tag named 'latest' is pushed
+# Note. In the last case, the tag can be (re-)created using the Git CLI, e.g.:
+#    git push -f origin <abfb967>:refs/tags/latest
```
gunzip, line 23: outdated

.devops/deploy-pipelines.yml
```diff
@@ -0,0 +1,152 @@
...
+              deploymentMethod: 'auto'
+              deployToSlotOrASE: true
+              slotName: 'staging'
+            displayName: Deploy to staging slot
+           # Second step: check the app is working
+          - template: templates/rest-healthcheck/template.yaml@pagopaCommons 
+            parameters:
+              azureSubscription: '$(PRODUCTION_AZURE_SUBSCRIPTION)'
+              appName: '$(PRODUCTION_APP_NAME)'
+              endpoint: 'https://$(PRODUCTION_APP_NAME)-staging.azurewebsites.net/info'
```
gunzip, line 138: isn't /api/v1/info? 
isn't it better to take only the path (/api/v1/info) from an env? or take the path + the URL with
two default values (https://$(PRODUCTION_APP_NAME)-staging.azurewebsites.net for the URL, /info for the path)

gunzip (changes requested):

.devops/code-review-pipelines.yml
```diff
@@ -0,0 +1,106 @@
...
+#    git push -f origin <abfb967>:refs/tags/latest
+trigger: none
+
+# This pipeline has been implemented to be run on hosted agent pools based both
+# on 'windows' and 'ubuntu' virtual machine images and using the scripts defined
+# in the package.json file. Since we are deploying on Azure functions on Windows
+# runtime, the pipeline is currently configured to use a Windows hosted image for
+# the build and deploy.
+pool:
+  vmImage: 'windows-2019'
```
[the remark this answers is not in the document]
balanza: I think it's important to have on Windows as it's the environment on which it would been deployed at last

.devops/deploy-pipelines.yml
balanza: I'd take the whole endpoint as it isn't so much an effort to specify, but it may be hard to abstract for other cases

balanza: The correct path should be `/api/info`

.devops/code-review-pipelines.yml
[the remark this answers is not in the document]
balanza: I inverted the default: now we use `ubuntu` as default and `windows` specifically for building. Same behaviour, but maybe more eloquent

.devops/deploy-pipelines.yml
gunzip: > I'd take the whole endpoint as it isn't so much an effort to specify, but it may be hard to abstract for other cases

is this the only parameter that differs between functions azure-template.yml? (it looks like everything else is the same for every function)

PR #99, opened 2020-12-03
[#175476527] Refactor pipeline

balanza:
Add:
* use a setup template 
* fix hit cache variable
* update template version


gunzip (approved):

gunzip:
it looks like the build step canno find npm-run-all command, is this intended?

pasqualedevita (changes requested):
same as https://github.com/pagopa/io-functions-app/pull/123#pullrequestreview-543861392
healthcheck ok

balanza:
> it looks like the build step cannot find npm-run-all command, is this intended?

It happens by mistake because `yarn install` is not executed

PR #106, opened 2021-02-17
Upgrate to @pagopa/openapi-codegen-ts

balanza:
Upgrade from `italia-utils` to `@pagopa/openapi-codegen-ts`


BurnedMarshal (approved):
lgtm

PR #109, opened 2021-03-30
[#IP-84] Change strict configuration to true

fabriziopapi:
Set the stirct configuration to true.
Fixed the compilation errors caused by the new settings.
The loigc/functions has not been rewritten or refactored: properly type cast has been introduced in order to pass the strict compiler check.

### DIFF
--- a/CreateMessage/__tests__/handler.test.ts
+++ b/CreateMessage/__tests__/handler.test.ts
@@ -179,7 +179,7 @@ describe("createMessageDocument", () => {
 
           expect(mockMessageModel.create).toHaveBeenCalledTimes(1);
           expect(response.isRight()).toBeTruthy();
-          expect(response.getOrElse(undefined)).toMatchObject({
+          expect(response.fold(l=>undefined,r=>r)).toMatchObject({
             fiscalCode,
             id: messageId,
             indexedId: messageId,
@@ -242,7 +242,7 @@ describe("forkOrchestrator", () => {
               serviceVersion: service.version
             })
           );
-          expect(response.getOrElse(undefined)).toEqual("orchestratorId");
+          expect(response.fold(l=>undefined,r=>r)).toEqual("orchestratorId");
         }
       )
     );
--- a/CreateMessage/handler.ts
+++ b/CreateMessage/handler.ts
@@ -376,7 +376,7 @@ export function CreateMessageHandler(
       telemetryClient.trackEvent({
         name: "api.messages.create",
         properties: {
-          error: isSuccess ? undefined : r.kind,
+          error: isSuccess ? "" : r.kind,
           hasDefaultEmail: Boolean(
             messagePayload.default_addresses &&
               messagePayload.default_addresses.email
--- a/CreateService/handler.ts
+++ b/CreateService/handler.ts
@@ -50,6 +50,7 @@ import {
 } from "italia-ts-commons/lib/strings";
 import { APIClient } from "../clients/admin";
 import { Service } from "../generated/api-admin/Service";
+import { ServiceMetadata } from "../generated/api-admin/ServiceMetadata";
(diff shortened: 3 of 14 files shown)


balanza (changes requested):
Manual casting must be considered the very very last resort. Every time we do that, we are poisoning our type system. It may happen once, but usually it's just problem-shifting - or worse: problem-hiding.

We must use alternatives; not that easy, though.

balanza:
I think we cannot avoid to manual cast things around, at the moment. We can group casts by their source problem into 2 groups:

**1. [withApiRequestWrapper](https://github.com/pagopa/io-functions-services/blob/442f196ea8b7e3293db051ea9af570bdd8a1df92/utils/api.ts#L16) method fails to narrow the type `T`**
This happens because the type T is inferred from the type of `apiCallWithParams`
```ts
Promise<t.Validation<IResponseType<number, T, never>>>
```
which is not correct, as the payload might be not only `T` but also other types. A better definition would probably be
```ts
Promise<
    | t.Validation<IResponseType<C, T, never>>
    | t.Validation<IResponseType<number, ErrorResponses, never>>
  >
```
where `C` is a subset of `number` which defines our acceptance status (eg: `200`).
The problem with the latter definition is that we're no longer able to tell `T` from `T | ErrorResponses` - or better: we never were able, but we didn't notice as we weren't on strict mode.
Me and @fabriziopapi spent a morning trying to refactor such functions, but we failed.
The simplest workaround would be to just get rid of it (its only purpose is to avoid too much boilerplate code due to type lifting. 

**2. `Service` entity is modeled loosey**
`Service` model is defined starting from `ServicePayload`, which is the DTO we expect from our API's users. To allow partial payloads (that is: you don't need to provide all fields until the service is _visible_), `ServicePayload` is defined by the union of `VisibleServicePayload` and `HiddenServicePayload`, with the latter having `service_metadata` nullable. This has been introduced after the discussion in https://github.com/pagopa/io-functions-services/pull/85.

As a result, api operations such as [createService](https://github.com/pagopa/io-functions-services/blob/IP-84--enforce-strict-to-true/createservice/handler.ts#L134) expect either an empty or fully-defined `service_metadata` field, but as we're reading such field from a given `ServicePayload` we simply cannot do it, so we arrange an object and we manually cast it as `service_metadata` like in https://github.com/pagopa/io-functions-services/blob/IP-84--enforce-strict-to-true/createservice/handler.ts#L141.
 
The issue is the field `token_name` into `service_metadata` must never be null, and that conflicts with the above definition; so we have no strong solution other than remodel `Service` data type.

---

Given the issue above, I think we should abandon this task and wait until we have them solved.

balanza:
> 2. `Service` entity is modeled loosey

Addendum: one issue is `token_name` is defined into `service_metadata` but both can be valued and/or null for independent reasons (that is: `token_name` can have a value while `service_metadata` is null). That conflicts with our model which includes the first inside the latter, so that if we have `token_name` defined and `service_metadata` undefined, we have to patch a dirty `ServiceMetadata` object just to include the token. Such operation can be only made with a manual cast.

Are we experiencing production issues? Not at the moment.
The chart below shows the failure count happening on `CreateService` endpoint exposed by `fn-admin` (i.e. the one we call from this very application, [here](https://github.com/pagopa/io-functions-services/blob/IP-84--enforce-strict-to-true/CreateService/handler.ts#L133) for example).
By just reading the model, I would expect some `400` due to the fact that we are _eventually_ sending a `ServiceMetadata` without a valued `scope` field; instead, production traffic prove me wrong. 

Other said: the case in which we receive an empty `service_metadata`, even for hidden Services, just never happens.

I don't have enough data to state if it is just contingency and if such event may happen.

Screenshot 2021-04-08 at 17 08 25

PR #112, opened 2021-04-06
Dependency check

balanza:
Add a check on dependency after the build to ensure that every imported module belongs to a package which has been explicitly declared as dependency.

This is a static check to avoid the involuntary use of a dev dependency in production code


BurnedMarshal (approved):
LGTM

balanza:
`fast-check` is actually needed by a test utility, `utils/arbitraries.ts`. I needed to explicitly mark such module as a test util, by doing:
* move it into a `__tests__` folder
* change jest config to only consider files that both
  * are into a `__tests__` folder
  * end with `.test.ts`

PR #117, opened 2021-05-04
[#IP-164] Bundled OpenAPI spec

balanza:
Expose an OpenAPI specification file which is bundled with its dependencies, which are no longer referenced as url to static file on another Github repo.

`openapi/index.yaml` file is now meant to be a generated file, having `openapi.ì/index.yaml.template` as template.
Generation occurs before the build.
A check on the code review pipeline is added so that we ensure changes to definitions are reflected into the exposed specification file.


AleDore (approved):
LGTM

gquadrati (approved):
Very useful, thanks!

PR #125, opened 2021-05-26
[#IP-210] Add 403 to API specs where missing

balanza:
Several APIs are exposed without explicitly declaring a 403 can happen, although they actually do. Mostly, this status can happen when the DevPortal user doesn't belong to the group allowed for such operaration.

Operations are: 

- `submitMessageforUserWithFiscalCodeInBody`
- `submitMessageforUser`
- `getMessage`
- `getProfileByPOST`
- `getProfile`
- `getSubscriptionsFeedForDate`
- `createService`
- `getUserServices`

### DIFF
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -157,6 +157,8 @@ paths:
           examples: {}
         '401':
           description: Unauthorized
+        '403':
+          description: Forbidden.
         '429':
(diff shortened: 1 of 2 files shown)


BurnedMarshal (approved):
LGTM

PR #128, opened 2021-06-07
[#IPV-22] Enable deploy on multiple app services

balanza:
#### List of Changes

* edit deploy pipeline to allow parallel deploy on multiple app services
* update pipelines templates

#### Motivation and Context

We may need to increase batch job parallelism. As they are built using Azure Durable Functions, and as they have an upper bound of maximum number of Orchestrators running at the same time, to increase such value we must deploy on different app services at once.

So far, the deploy pipeline still has a default of one app service (the one actually deployed). We may change this value in the future, still we can provide different inputs before running the pipeline.

#### Types of changes

#### Checklist:

### DIFF
--- a/.devops/deploy-pipelines.yml
+++ b/.devops/deploy-pipelines.yml
@@ -1,9 +1,5 @@
 # Azure DevOps pipeline to release a new version and deploy to production.
 
-variables:
-  NODE_VERSION: '10.14.1'
-  YARN_CACHE_FOLDER: $(Pipeline.Workspace)/.yarn
-
 parameters:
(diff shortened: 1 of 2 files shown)


.devops/deploy-pipelines.yml
```diff
@@ -31,7 +48,7 @@ resources:
     - repository: pagopaCommons
       type: github
       name: pagopa/azure-pipeline-templates
-      ref: refs/tags/v14
+      ref: refs/tags/v10
```
balanza, line 51:
```suggestion
ref: refs/tags/v14
```

PR #131, opened 2021-06-10
[#IP-234] Disable message content for WEBHOOK channel

BurnedMarshal:
#### List of Changes
- Add new feature flag `FF_DISABLE_WEBHOOK_MESSAGE_CONTENT` in configuration (default `false`)
- With `FF_DISABLE_WEBHOOK_MESSAGE_CONTENT` enabled all the webhook message content will be disabled

#### Motivation and Context
We want send push notifications with a generic payload and title. With `FF_DISABLE_WEBHOOK_MESSAGE_CONTENT` enabled we remove content from the webhook message, and `io-backend` provide a generic one.

unit test

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

#### Checklist:

### DIFF
--- a/WebhookNotificationActivity/__tests__/handler.test.ts
+++ b/WebhookNotificationActivity/__tests__/handler.test.ts
@@ -5,7 +5,7 @@
 jest.mock("applicationinsights");
 jest.mock("azure-storage");
 
-import { isLeft, left, right } from "fp-ts/lib/Either";
+import { isLeft, isRight, right } from "fp-ts/lib/Either";
 import { none, some } from "fp-ts/lib/Option";
 import {
   EmailString,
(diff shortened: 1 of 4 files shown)


balanza:
What's the point of having a feature flag switch? Is this a feature we want to enable/disable over time?

If not, maybe we can just get rid of conditions and send no content by default

BurnedMarshal:
> What's the point of having a feature flag switch? Is this a feature we want to enable/disable over time?
> 
> If not, maybe we can just get rid of conditions and send no content by default

Maybe we want to re-enable content in push notification when we'll have:
1. opt-in push notification disclaimer in app
2. end-to-end push notification encryption

pasqualedevita (approved):

pasqualedevita:
Can we merge?

balanza (approved):

PR #142, opened 2021-07-07
[#IP-275] Managed service preferences before message creation

michaeldisaro:
#### List of Changes
- Added checks for user's services preferences during the message creation activity

#### Motivation and Context
We want to create messages only for users that allow to be reached, and return a coherent blockedInboxOrChannels based on preferences if they are present.

It has been tested by performing:
- yarn install --frozen-lockfile
- yarn build
- yarn lint
- yarn test

#### Types of changes

#### Checklist:

- [x] My change requires a change to the documentation.


gquadrati (approved):
LGTM

StoreMessageContentActivity/handler.ts
```diff
@@ -135,62 +234,100 @@ export const getStoreMessageContentActivityHandler = (
...
-  // visible to getMessages by changing the pending flag to false
-  const updatedMessageOrError = await lMessageModel
-    .upsert({
-      ...newMessageWithoutContent,
-      isPending: false
-    })
-    .run();
+      return blockedInboxOrChannels;
+    }, identity)
+    .map<Promise<StoreMessageContentActivityResult>>(
```
AleDore, line 271: Why not chain this with subsequent tasks in a more functional way. I'm wondering about manage Promise using `await` inside map operation, while having the possibility to use `chain` capability. 
AFAIK, in alternative we can mantain a more procedural style without using `chain` construct as  the previous version :). What do you think?

michaeldisaro: To comply with new requests I decided to keep the old "good" code as a "block" to use where it was necessary, I can take a look to transform the old code to a new and more functional one. I'll let you know.

StoreMessageContentActivity/handler.ts
```diff
@@ -53,13 +67,107 @@ export type StoreMessageContentActivityResult = t.TypeOf<
...
+) => ReadonlyArray<BlockedInboxOrChannelEnum> = servicePreference =>
+  Object.entries(servicePreference)
+    // take only attributes of ServicePreferencesValues
+    .filter(([name, _]) => channelToBlockedInboxOrChannelEnum[name])
+    // take values set to false
+    .filter(([_, isEnabled]) => !isEnabled)
+    // map to BlockedInboxOrChannelEnum
+    .map(([name, _]) => channelToBlockedInboxOrChannelEnum[name]);
+
+const getServicePreferenceValueOrError = (
```
balanza, line 111: If I get it right, the purpose of this subroutine is to reduce the complexity of User's preferences into an array of Services to be blocked. Right?

It would be useful to have a comment which explains that, so that readers do not have to go through inner, technical detail to figure out what the function does.

StoreMessageContentActivity/handler.ts
```diff
@@ -53,13 +67,107 @@ export type StoreMessageContentActivityResult = t.TypeOf<
...
+    )
+    .map(maybeServicePref =>
+      maybeServicePref.foldL<ReadonlyArray<BlockedInboxOrChannelEnum>>(
+        () =>
+          // if we do not have a preference we return an empty array only
+          // if we have preference mode AUTO, else we must return an array
+          // with BlockedInboxOrChannelEnum.INBOX
+          userServicePreferencesMode === ServicesPreferencesModeEnum.AUTO
+            ? []
+            : [BlockedInboxOrChannelEnum.INBOX],
```
balanza, line 158: This is ok so far, but what if we add another way to handle subscription for our users (that is: we add another case to `ServicesPreferencesModeEnum`)? Such new case would be considered in the _else_ case, just like `MANUAL`, and we wouldn't notice that.

My advice is to always be explicit when handling enums, and treat unexpected values for what they are: values for which we cannot make decisions.

[Here](https://github.com/pagopa/io-functions-services/blob/2df422b7ba2e582f0a4b8e4030a0b6a854b3c154/utils/profile.ts#L116) you can see an example

StoreMessageContentActivity/handler.ts
```diff
@@ -53,13 +67,107 @@ export type StoreMessageContentActivityResult = t.TypeOf<
...
+        userServicePreferencesVersion as NonNegativeInteger
+      )
+    )
+    .chain(documentId =>
+      servicePreferencesModel
+        .find([documentId, fiscalCode])
+        .mapLeft<ServicePreferenceError>(failure => ({
+          kind: "ERROR",
+          message: `COSMOSDB|ERROR=${failure.kind}`
+        }))
```
balanza, line 148: Instead of wrapping `CosmosErrors` into `ServicePreferenceError` type, maybe it's more eloquent to have them in union:
```ts
// the following it's just an example, you got the idea
type InvalidModeError = {
  kind: "INVALID_MODE", // what you call "LEGACY"
  value: ServicesPreferencesModeEnum
}

type ServicePreferenceError = InvalidModeError | CosmosErrors
```

Benefits are:
* you do not have manually re-map `CosmosErrors`
* the signature is more explicit about what kind of failure can happen



I made something similar [here](https://github.com/pagopa/io-functions-services/blob/2df422b7ba2e582f0a4b8e4030a0b6a854b3c154/utils/profile.ts#L104)

StoreMessageContentActivity/handler.ts
balanza: > I decided to keep the old "good" code as a "block"

I think this is a good design decision. 

> a new and more functional one

That would require a huge refactor, are you sure it is worth the effort?

I think we can have best of both world by doing the following:
* write new code using FP, and isolate that in a well-scoped function (you did it already in `getServicePreferenceValueOrError`)
* use such function inside the _legacy_ imperative code

Something like:
```ts
const remappedBlockedInboxOrChannels = 
    await getServicePreferenceValueOrError(/* whatever */)
		.fold(/* in case you need it */)
		.run();

// whether the user has blocked inbox storage for messages from this sender
const isMessageStorageBlockedForService =
	remappedBlockedInboxOrChannels.indexOf(
		BlockedInboxOrChannelEnum.INBOX
	) >= 0;
	
// ...and code follows from here...
```	
	
	
@AleDore @michaeldisaro what do you think?

michaeldisaro: I'm just trying to do that refactor, if I'll close it for this afternoon without test errors it could be a good chance to have better functional code. At least I tried...

StoreMessageContentActivity/handler.ts
michaeldisaro: Seems legit.

StoreMessageContentActivity/handler.ts
michaeldisaro: I check.

StoreMessageContentActivity/handler.ts
michaeldisaro: Seems legit.

StoreMessageContentActivity/handler.ts
gquadrati: This is my _reposability_, actually :D
and yes, it's exactly what you wrote; rather than writing 3 `if` to check the enabled channels, we decided to do like this. So in case a new channel is added, missing mapping will break the build

StoreMessageContentActivity/handler.ts
```diff
@@ -53,13 +67,107 @@ export type StoreMessageContentActivityResult = t.TypeOf<
...
+  isEmailEnabled: BlockedInboxOrChannelEnum.EMAIL,
+  isInboxEnabled: BlockedInboxOrChannelEnum.INBOX,
+  isWebhookEnabled: BlockedInboxOrChannelEnum.WEBHOOK
+};
+
+const servicePreferenceToBlockedInboxOrChannels: (
+  servicePreference: ServicePreference
+) => ReadonlyArray<BlockedInboxOrChannelEnum> = servicePreference =>
+  Object.entries(servicePreference)
+    // take only attributes of ServicePreferencesValues
```
gquadrati, line 104:
```suggestion
) => ReadonlyArray<BlockedInboxOrChannelEnum> = servicePreference =>
  /**
   * Reduce the complexity of User's preferences into an array of BlockedInboxOrChannelEnum
   * In case a preference is set to false, it is translated to proper BlockedInboxOrChannelEnum
   * and added to returned array.
   * By adding a `channelToBlockedInboxOrChannelEnum` map we are prepared to handle new
   * service preferences
   */
  Object.entries(servicePreference)
    // take only attributes of ServicePreferencesValues
```

StoreMessageContentActivity/handler.ts
balanza: Rule of thumb: when adding a function, comment **why** you need it

StoreMessageContentActivity/__tests__/handler.test.ts
```diff
@@ -253,11 +296,1075 @@ describe("getStoreMessageContentActivityHandler", () => {
...
+    );
+    expect(findServicePreferenceMock).toHaveBeenCalled();
+    expect(findServicePreferenceMock).toHaveBeenCalledTimes(1);
+    expect(findServicePreferenceMock).toHaveBeenCalledWith([
+      documentId,
+      aRetrievedProfile.fiscalCode
+    ]);
+  });
+
+  it("should return a failure if message sender service exists in profile.blockedInboxOrChannels (LEGACY SETTINGS)", async () => {
```
BurnedMarshal, line 1225: is this the same scenario tested in [here](https://github.com/pagopa/io-functions-services/pull/142/files?diff=split&w=1#diff-3d5477f69edf2d4434f5f726aef090d871dffe3b10040fb4e72a62da72cb2e8fR510)?

StoreMessageContentActivity/__tests__/handler.test.ts
```diff
@@ -253,11 +296,1075 @@ describe("getStoreMessageContentActivityHandler", () => {
...
+    // => FAILURE
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalled();
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalledTimes(1);
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalledWith([
+      aRetrievedProfile.fiscalCode
+    ]);
+    expect(findServicePreferenceMock).not.toHaveBeenCalled();
+  });
+
+  it("should return a success if message sender service does not exists in profile.blockedInboxOrChannels (LEGACY SETTINGS)", async () => {
```
BurnedMarshal, line 1273: seems even this test was tested before

StoreMessageContentActivity/__tests__/handler.test.ts
```diff
@@ -253,11 +296,1075 @@ describe("getStoreMessageContentActivityHandler", () => {
...
+    // => SUCCESS
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalled();
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalledTimes(1);
+    expect(findLastProfileVersionByModelIdMock).toHaveBeenCalledWith([
+      aRetrievedProfile.fiscalCode
+    ]);
+    expect(findServicePreferenceMock).not.toHaveBeenCalled();
+  });
+
+  it("should return a success if message sender service exists in profile.blockedInboxOrChannels but only blocks EMAIL (LEGACY SETTINGS)", async () => {
```
BurnedMarshal, line 1321: ditto

BurnedMarshal (changes requested):

StoreMessageContentActivity/handler.ts
michaeldisaro: Present in https://github.com/pagopa/io-functions-services/pull/142/commits/b50a4f98a99bbfe99da4326619f7b1434660e716

StoreMessageContentActivity/handler.ts
```diff
@@ -135,62 +241,100 @@ export const getStoreMessageContentActivityHandler = (
...
+          blockedInboxOrChannels
+        )}`
+      );
 
-  if (isLeft(errorOrAttachment)) {
-    context.log.error(`${logPrefix}|ERROR=${errorOrAttachment.value}`);
-    throw new Error("Error while storing message content");
-  }
+      // an error occurs also when user service preference mode is LEGACY
+      context.log.warn(`${logPrefix}|${servicePreferenceError.message}`);
```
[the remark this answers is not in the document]
michaeldisaro: I thought the same, I already removed it in https://github.com/pagopa/io-functions-services/pull/142/commits/b50a4f98a99bbfe99da4326619f7b1434660e716

StoreMessageContentActivity/handler.ts
michaeldisaro: Resolved in https://github.com/pagopa/io-functions-services/pull/142/commits/b50a4f98a99bbfe99da4326619f7b1434660e716

StoreMessageContentActivity/__tests__/handler.test.ts
michaeldisaro: Clicking the link brings me to the "changed files" section. I cannot see which test you refer to.

michaeldisaro: Tomorrow I will check LEGACY tests. Initially they started different, but it's feasible that some have converged to the same test during development.

BurnedMarshal: Yes, no problem ;)
Thanks

michaeldisaro: Removed in https://github.com/pagopa/io-functions-services/pull/142/commits/b9d58101e54264cc46f6fa6221efdc7febeba3e8

StoreMessageContentActivity/__tests__/handler.test.ts
michaeldisaro: Removed in https://github.com/pagopa/io-functions-services/pull/142/commits/b9d58101e54264cc46f6fa6221efdc7febeba3e8

StoreMessageContentActivity/__tests__/handler.test.ts
michaeldisaro: Removed in https://github.com/pagopa/io-functions-services/pull/142/commits/b9d58101e54264cc46f6fa6221efdc7febeba3e8

StoreMessageContentActivity/handler.ts
michaeldisaro: @balanza I made some refactoring, not as deep as @AleDore suggested, in https://github.com/pagopa/io-functions-services/pull/142/commits/b232fc7bbe7952e787c06e39f0450557e671ae0e and previous commits.

StoreMessageContentActivity/handler.ts
```diff
@@ -53,13 +73,206 @@ export type StoreMessageContentActivityResult = t.TypeOf<
...
+  Object.entries(servicePreference)
+    // take only attributes of ServicePreferencesValues
+    .filter(([name, _]) => channelToBlockedInboxOrChannelEnum[name])
+    // take values set to false
+    .filter(([_, isEnabled]) => !isEnabled)
+    // map to BlockedInboxOrChannelEnum
+    .map(([name, _]) => channelToBlockedInboxOrChannelEnum[name]);
+
+/**
+ * Converts a preferemce to a remapped blockedInboxOrChannels if it exists
```
balanza, line 155: typo:
```suggestion
 * Converts a preference to a remapped blockedInboxOrChannels if it exists
```

balanza:
The code seems to work as expected, anyway by a reviewer perspective is hard for me to focus on differences because the whole activity has been refactored. I'm afraid I cannot spot eventual bug/regression/edge-case we may have included in those changes.

To help on this I rewrote the test suite so that is easier to figure out which scenario we're testing and which we're not. So far I just replicated tests we used to have, and group such in three families: successes, failures and exceptions.



Changes are in https://github.com/pagopa/io-functions-services/pull/142/commits/7e98711c942834885d2abf6f9b561422cf9fab85

michaeldisaro:
> The code seems to work as expected, anyway by a reviewer perspective is hard for me to focus on differences because the whole activity has been refactored. I'm afraid I cannot spot eventual bug/regression/edge-case we may have included in those changes.
> 
> To help on this I rewrote the test suite so that is easier to figure out which scenario we're testing and which we're not. So far I just replicated tests we used to have, and group such in three families: successes, failures and exceptions.
> 
> Changes are in [7e98711](https://github.com/pagopa/io-functions-services/commit/7e98711c942834885d2abf6f9b561422cf9fab85)

Refactoring happened mainly to avoid linting errors due to too much parameters in function signature. Other refactoring happened to extract the old code and make the logic more understandable. 

Your test refactoring seems legit.

BurnedMarshal (approved):
LGTM

PR #144, opened 2021-07-08
[#IP-301] `GetSubscriptionFeed` Add service subscription check on profile unsubscription

AleDore:
User that enables a single service after switching to MANUAL mode, must be tracked on Subscription Feed because Profile Unsubscription is not related only to a Profile deletion.

#### List of Changes

- `GetSubscriptioFeed` evaluate service subscription while it reads a profile unsubscription
- Update test

#### Motivation and Context

It solves a data inconsistence if User enables a single service after switching to MANUAL mode in the same day.

Unit tested
#### Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

#### Checklist:


BurnedMarshal (approved):
LGTM

PR #146, opened 2021-07-13
[#IP-303] Add tracing for message storing

balanza:
Add the `api.messages.create.blockedstoremessage` custom event so that we can monitor:
* how many messages are accepted/rejected due to Citizen preferences
* how many Citizens are receiving messages and are in `LEGACY` mode
* how many messages are rejected, grouped by preference mode


BurnedMarshal (approved):
LGTM

PR #147, opened 2021-07-14
[#IP-303] Add tracing for `GetLimitedProfileByPOST` and `GetLimitedProfile`

balanza:
Add `api.limitedprofile.sender-allowed` custom event that traces the response of `GetLimitedProfileByPOST` and `GetLimitedProfile` based on Citizen's preference mode, so that we can monitor:
* how many profiles are asked and are still in `LEGACY` mode
* how many times a sender is not allowed to send to a Citizen it would like to, grouped by mode


utils/profile.ts
```diff
@@ -237,19 +240,33 @@ export const getLimitedProfileTask = (
...
+                  error
+                ),
+          isAllowed => ({
+            isAllowed,
+            profile
+          })
+        )
+        .map(_ => {
+          telemetryClient.trackEvent({
+            name: "api.messages.subscription.senderallowed",
```
BurnedMarshal, line 259: What if we rename the event name related to the specific api. For example `api.limitedprofile.sender-allowed`

BurnedMarshal (approved):
LGTM, only one question about the event name

utils/profile.ts
balanza: Fine for me, I struggled a bit about the name

PR #167, opened 2021-11-13
[IC-90] Add integration tests

gquadrati:
#### Types of changes

#### Checklist:


__integrations__/index.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+beforeAll(async () => {});
+afterAll(async () => {});
+
+beforeEach(() => jest.clearAllMocks());
+
+const aSubscriptionKey = "aSubscriptionKey";
+const aContent = aMessageContent;
+
+describe("fn-services |> ping", () => {
```
gunzip, line 23: I'm not sure about the added value of this kind of integration tests. 
They require a lot of setup and moving parts. What are we testing here?
Under which conditions a unit tested endpoint could return a result different to what's expected?

gquadrati: I'm just setting up the environment, then I'll add message creation tests. 
We are going to refactor that process, so we need to test it

balanza: > [...]They require a lot of setup and moving parts [...] Under which conditions a unit tested endpoint could return a result different to what's expected?

@gunzip the value is to ensure all moving parts are cooperating properly. We don't expect such kind of integration tests to cover all the spectrum of possible cases, we just want to ensure a workflow which involves more than a function can be processed as a whole.

balanza:
Great job @gquadrati!

I see tests aren't working in CI/CD environment, but they are when executed locally. That means a misconfiguration of some kind, let just not spend too much time on it for the moment. 
To have a replicable integration test suite is already a good achievement, I'm sure we can fix pipeline soon after.

If you agree, I propose:
* we remove the stage from the pipeline
* we merge this into master

gquadrati:
> Great job @gquadrati!
> 
> I see tests aren't working in CI/CD environment, but they are when executed locally. That means a misconfiguration of some kind, let just not spend too much time on it for the moment. To have a replicable integration test suite is already a good achievement, I'm sure we can fix pipeline soon after.
> 
> If you agree, I propose:
> 
> * we remove the stage from the pipeline
> * we merge this into master

Thanks! 
But I finally made them work, so we can merge it

__integrations__/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "version": "1.0.0",
+  "description": "Integration test suite for integration testing",
+  "license": "MIT",
+  "keywords": [],
+  "author": "",
+  "scripts": {
+    "prestart": "npm-run-all generate-env",
+    "start": "docker-compose --env-file environments/env.base up -d",
+    "stop": "docker-compose --env-file environments/env.base down",
+    "stop:all": "docker-compose --env-file environments/env.base down && docker image prune -a -f && docker volume prune -f",
```
balanza, line 12: > docker image prune -a -f && docker volume prune -f

Not actually useful for local development (every successive run will need to fetch all layers from registry) nor for CI/CD (VMs are managed and I don't think we should care too much).

I suggest to remove it

```suggestion
    "stop:all": "docker-compose --env-file environments/env.base down",
```

__integrations__/index.test.ts
```diff
@@ -0,0 +1,406 @@
...
+      break;
+    } catch (e) {
+      await delay(WAIT_MS);
+      i++;
+    }
+  }
+  if (i >= 10) {
+    console.log("Function unable to setup in time");
+    exit(1);
+  }
```
balanza, line 118: I see this whole chunk of code is to handle function app startup time - which is something that should not be a responsibility of the test suite itself: tests should assume the environment is up&running.

I would not point that out if I didn't spot a flaw myself: if something is wrong, or even tests fail in unexpected manner, the while loop will continue to run even if Jest stopped, resulting in a flood of error message like 

> Cannot log after tests are done. Did you forget to wait for something async in your test?

I'd rather have a delay big enough (even 30s is acceptable to me): even in case it would not suffice, we would have a clear error reporting. Also, we can choose to use `sleep` bash command:
```sh
yarn start
sleep 30s
yarn test
```

It's not mandatory, though

balanza: UPDATE: I found how to replicate my error (and why we are exposed to the flaw).

I was starting the test suit bad, thus my test agent wasn't able to connect to the function app, ever - thus iterating 100 times with 10s delay each.

Having Jest a default timeout of 25s on hooks execution, after the third failed attempt the worked ended, leaving 97 iterations to fail because I was logging after the test ended 🙃 .

__integrations__/docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+      - languageWorkers__node__arguments="--inspect=0.0.0.0:5861"
+    links:
+      - cosmosdb
+      - storage-account
+    depends_on:
+      - functions-node-14
+      - fixtures
+
+  testagent:
+    image: node:16-alpine
```
michaeldisaro, line 66: why node 16?

__integrations__/package.json
gquadrati: This is already present as "yarn stop" command

__integrations__/index.test.ts
gquadrati: I know this is not good, but I was not able to speed up function container setup in code review pipeline. 
maybe I can set the maximum retry to 25 (and actually check the correct value -.-")

I also wanted to move to local .env file so we can decrease the default timeout

__integrations__/index.test.ts
```diff
@@ -0,0 +1,406 @@
...
+});
+
+describe("fn-services |> ping", () => {
+  it("/info return 200", async () => {
```
michaeldisaro, line 122: Test is against 200 and we expect 500?

michaeldisaro (approved):
Apart some comments, code looks good.

__integrations__/index.test.ts
balanza: I stick with my opinion: big enough `sleep` :)

__integrations__/docker-compose.yml
balanza: I don't know if it's intended, anyway tests are _against_ function app, not _on_ it, so we're not constrained to use the very same runtime version.

__integrations__/index.test.ts
gquadrati: Thanks, I remove it, since it was not necessary

__integrations__/index.test.ts
gquadrati: I refactor it a bit in [951e824](https://github.com/pagopa/io-functions-services/pull/167/commits/951e824e57f37152291038145b0d766aa764cd5c)

PR #169, opened 2021-11-14
[#IC-87] Dismantle durable functions

balanza:
#### Motivation and Context

Following #168 we remove unused code

#### Types of changes

#### Checklist:


BurnedMarshal (approved):
LGTM

gunzip:
For [REDACTED] project we need to collect several metrics that can tell us how much time is elapsed between a message reaches the `io-function-services` HTTP controller and: 

1. the moment it becomes visible in the user inbox (I guess, when it's stored into cosmosdb + storage account and the status switches to "processed")
1. the moment the webhook is called

Ideally: the time spent in each step of the message path.

Merely this rector cuts out some [formerly tracked events](https://github.com/pagopa/io-functions-services/blob/v1.17.1/CreatedMessageOrchestrator/utils.ts) that could have helped here.

This need is due to the reason we probably need some **priority mechanism** to deliver the messages that come from some "critical" services, but currently we don't know the average time needed to deliver one message (inbox + push, email does not matter atm). If the average time is under, lets say 10 seconds, even under peak loads, probably we won't need a refactor at all.

I've read the [IO-RFC with the analysis of probable bottlenecks](https://docs.google.com/document/d/1DjZ-EFWNWgM8cHQ4gkcCN-c4g0Wp9vQfC7RmGnNdhVM/edit#heading=h.hiv7ixk5t7oh) and, as far as I understood, the main ones are storage queues (Am I missing something @BurnedMarshal ?).

Another option, to get some insights, could be to monitor the length of the queues which are used currently.

What do you think about that? 

@BurnedMarshal @balanza

balanza:
> Merely this rector cuts out some formerly tracked events

We can reintroduce such events with little effort. Please keep in mind that such events are (and were) sampled, so that you can easily quantify the % of processed/failing/whatever messages, but you cannot trace a single message through the entire flow. 

What we might want to achieve is a _transactional tracing_ of some form, so that we can follow (every? some?) messages in each step from the sender to the Citizen's inbox.

gunzip:
> keep in mind that such events are (and were) sampled

No they werent' as a specific flag was introduces to avoid sampling those events.

Anyway, it's ok even if they're sampled since what we want here is the average timing.

balanza:
> No they werent' as a specific flag was introduced by @infantesimone to avoid sampling those events.

AFAIK we had empiric evidence that sampling was applied anyway, but I may be wrong.

pasqualedevita:
I think is possibile (not simple) but using app insight can be not accurate due to sampling and we don't know if u can disable sampling with custom events.
As alternative I want to propose to send relevant events in a kafka queue (eventhub). In this way u can read the queue and create all kpi that u need in PDND dashboards (filer by sender service, priority, average, percentile distribution, ecc)

PS: I think we are under 10s, probably now the bottleneck is only on fn-pushnotification

gunzip:
If you group by messageId sampling shouldn't matter since any of those two values would be simply undefined (you have to filter out undefined entries).

> I think we are under 10s, probably now the bottleneck is only on fn-pushnotification
great!

PR #172, opened 2021-11-22
[#IC-49] Porting SpecialServices to new ProcessMessage flow

BurnedMarshal:
#### List of Changes

Upgrade `io-functions-commons` dependency for SpecialServices
CreateMessage flow now have `serviceCategory` to handle Standard and Special service preferences check.
Only admin can create and edit `SpecialService` properties.

#### Motivation and Context

The `io-functions-services` should handle Standard and Special services. This PR upgrade specifications and models to do that.
When a message is created for a Special Service to evaluate if INBOX is enabled for a User, Activation model is used to retrieve Special Service activation.

unit test

#### Types of changes

- [X] New feature (non-breaking change which adds functionality)

### DIFF
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -455,6 +455,8 @@ export const getProcessMessageHandler = ({
                 createdMessageEvent.senderMetadata.serviceCategory ===
                 SpecialServiceCategoryEnum.SPECIAL
               ) {
+                // If the service is SPECIAL we need to retrieve the Activation information of that service
+                // and override the INBOX value inside the original blockInboxOrChannel
                 return getActivationForSpecialServices(lActivation)({
                   blockedInboxOrChannel,
                   context,
@@ -463,6 +465,7 @@ export const getProcessMessageHandler = ({
                   logPrefix
                 });
               }
+              // If the service is STANDARD we use the original value for blockedInboxOrChannel
               return TE.of(blockedInboxOrChannel);
             }),
             TE.toUnion
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -463,9 +463,6 @@ export const getProcessMessageHandler = ({
                 createdMessageEvent.senderMetadata.serviceCategory ===
                 SpecialServiceCategoryEnum.SPECIAL
               ) {
-                // If the service has category equals to SPECIAL, the INBOX value in blockedInboxOrChannel depends from the Activation
-                // value. If the SPECIAL service is ACTIVE (exists an activation for couple user/service with ACTIVE status)
-                // we remove INBOX from blockedInboxOrChannel, otherwise we add the INBOX value.
                 return getActivationForSpecialServices(lActivation)({
                   blockedInboxOrChannel,
                   context,
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -198,8 +198,10 @@ type ActivationForSpecialServices = (params: {
 >;
 
 /**
- * Override the INBOX value of BlockedInboxOrChannel with the Activation value
- * related to a Special Service and a User.
+ * Returns a new blockedInboxOrChannel value for a SPECIAL service related to a User.
+ * When a service has category SPECIAL the INBOX value depends from the activation status
+ * for the couple user/service.
+ * A missing Activation es equal an INACTIVE one.
  *
  * @param lActivation
  * @returns
@@ -455,8 +457,9 @@ export const getProcessMessageHandler = ({
                 createdMessageEvent.senderMetadata.serviceCategory ===
                 SpecialServiceCategoryEnum.SPECIAL
               ) {
-                // If the service is SPECIAL we need to retrieve the Activation information of that service
-                // and override the INBOX value inside the original blockInboxOrChannel
+                // If the service has category equals to SPECIAL, the INBOX value in blockedInboxOrChannel depends from the Activation
+                // value. If the SPECIAL service is ACTIVE (exists an activation for couple user/service with ACTIVE status)
+                // we remove INBOX from blockedInboxOrChannel, otherwise we add the INBOX value.
                 return getActivationForSpecialServices(lActivation)({
                   blockedInboxOrChannel,
                   context,
@@ -465,7 +468,8 @@ export const getProcessMessageHandler = ({
                   logPrefix
                 });
               }
-              // If the service is STANDARD we use the original value for blockedInboxOrChannel
+              // If the service is STANDARD we use the original value of blockedInboxOrChannel
+              // calculated from services preferences and user profile.
               return TE.of(blockedInboxOrChannel);
             }),
             TE.toUnion
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -31,6 +31,8 @@ import {
   MessageStatusModel
 } from "@pagopa/io-functions-commons/dist/src/models/message_status";
 import { MessageStatusValueEnum } from "@pagopa/io-functions-commons/dist/generated/definitions/MessageStatusValue";
+import { ActivationModel } from "@pagopa/io-functions-commons/dist/src/models/activation";
+import { ActivationStatusEnum } from "@pagopa/io-functions-commons/dist/generated/definitions/ActivationStatus";
 import { initTelemetryClient } from "../utils/appinsights";
 import { toHash } from "../utils/crypto";
 import { PaymentData } from "../generated/definitions/PaymentData";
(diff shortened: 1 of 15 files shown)


ProcessMessage/handler.ts
```diff
@@ -384,6 +389,49 @@ export const getProcessMessageHandler = ({
 
               return result;
             }),
+            TE.chain(_ => {
```
balanza, line 392: Consider move this sub-pipe into a dedicated function to increase redability and keep the main pipe at the same level of abstraction

UpdateService/handler.ts
```diff
@@ -137,10 +141,23 @@ const updateServiceTask = (
...
+          service_metadata: SpecialServiceMetadata.is(
+            retrievedService.service_metadata
+          )
+            ? ({
+                ...servicePayload.service_metadata,
+                category: SpecialServiceCategoryEnum.SPECIAL,
+                custom_special_flow:
+                  retrievedService.service_metadata.custom_special_flow,
+                token_name: adb2cTokenName
+              } as SpecialServiceMetadata)
```
balanza, line 154: Why the need for manual casting? I'm afraid it may hide some type inconsistency

openapi/index.yaml
```diff
@@ -756,7 +756,7 @@ definitions:
           created_at:
             $ref: '#/definitions/Timestamp'
           content:
-            $ref: '#/definitions/MessageContent'
+            $ref: '#/definitions/NewMessageContent'
```
balanza, line 759: I see this change to come from template rendering, I just wonder why it happened.

package.json
```diff
@@ -20,7 +20,7 @@
     "generate": "npm-run-all generate:*",
     "generate:definitions": "rimraf ./generated/definitions && shx mkdir -p ./generated/definitions && gen-api-models --api-spec ./openapi/index.yaml --no-strict --out-dir ./generated/definitions",
     "generate:api-notifications": "rimraf ./generated/notifications && shx mkdir -p ./generated/notifications && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-backend/master/api_notifications.yaml --out-dir ./generated/notifications --response-decoders --request-types",
-    "generate:api-admin": "rimraf generated/api-admin && shx mkdir -p generated/api-admin && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-functions-admin/master/openapi/index.yaml --no-strict --out-dir generated/api-admin --request-types --response-decoders --client",
+    "generate:api-admin": "rimraf generated/api-admin && shx mkdir -p generated/api-admin && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-functions-admin/IC-40-special-service-model-upgrade/openapi/index.yaml --no-strict --out-dir generated/api-admin --request-types --response-decoders --client",
```
balanza, line 23: We won't merge a specific branch. Consider:
* generate code from `master` (as it used to be)
* generate code from a specific version (good)
* import sdk package (better)

balanza (changes requested):

package.json
BurnedMarshal: You have right. This should be changed to `master` when the pagopa/io-functions-admin#172 is merged.
If we want to move to sdk import version I think is better do the refactory on another PR.

openapi/index.yaml
BurnedMarshal: This change comes from [here](https://github.com/pagopa/io-functions-commons/pull/218/files#diff-390ee41a01068fe2e94192fa8caf0b01f8ad5e5a71545ae89874cf453348b334R267) released with `v21.7.3` of `io-functions-commons`

UpdateService/handler.ts
BurnedMarshal: Good catch. All remove all casting. Fixed in 48003b8

ProcessMessage/handler.ts
gunzip: I'd also add a comment on what's happening

BurnedMarshal: Fixed in 8a9cfb5

BurnedMarshal: I've also added a comment in b2e90ed

BurnedMarshal:
> override with what? (why?) it's not very clear form the comment how this works

Sorry, I try to explain it better

BurnedMarshal:
> > override with what? (why?) it's not very clear form the comment how this works
> 
> Sorry, I try to explain it better

@gunzip I've fixed the comments 2ca3a1a. Tell me if now is more clear.

ProcessMessage/handler.ts
```diff
@@ -384,6 +452,26 @@ export const getProcessMessageHandler = ({
...
               return result;
             }),
+            TE.chain(blockedInboxOrChannel => {
+              if (
+                createdMessageEvent.senderMetadata.serviceCategory ===
+                SpecialServiceCategoryEnum.SPECIAL
+              ) {
+                // If the service has category equals to SPECIAL, the INBOX value in blockedInboxOrChannel depends from the Activation
+                // value. If the SPECIAL service is ACTIVE (exists an activation for couple user/service with ACTIVE status)
+                // we remove INBOX from blockedInboxOrChannel, otherwise we add the INBOX value.
```
gunzip, line 462:
```suggestion
```

ProcessMessage/handler.ts
```diff
@@ -384,6 +452,26 @@ export const getProcessMessageHandler = ({
...
+                return getActivationForSpecialServices(lActivation)({
```
gunzip, line 463:
```suggestion
                return getBlockedInboxesForSpecialService(lActivation)({
```

ProcessMessage/handler.ts
```diff
@@ -384,6 +452,26 @@ export const getProcessMessageHandler = ({
...
+                  blockedInboxOrChannel,
+                  context,
+                  createdMessageEvent,
```
gunzip, line 466: it looks like you only need the `senderServiceId`

ProcessMessage/handler.ts
```diff
@@ -183,6 +186,69 @@ const getServicePreferenceValueOrError = (
...
+}) => TaskEither<
+  ReadonlyArray<BlockedInboxOrChannelEnum>,
+  ReadonlyArray<BlockedInboxOrChannelEnum>
+>;
+
+/**
+ * Returns a new blockedInboxOrChannel value for a SPECIAL service related to a User.
+ * When a service has category SPECIAL the INBOX value depends from the activation status
+ * for the couple user/service.
+ * A missing Activation is equal to an INACTIVE one.
```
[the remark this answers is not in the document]
BurnedMarshal: Thanks a lot :D

ProcessMessage/handler.ts
```diff
@@ -183,6 +186,69 @@ const getServicePreferenceValueOrError = (
...
+ *
+ * @param lActivation
+ * @returns
```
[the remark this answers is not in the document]
BurnedMarshal: If you think that is more clear we can return a `TaskEither<never, ReadonlyArray<BlockedInboxOrChannelEnum>>` changing the last part of the chain like this:
```typescript
TE.chainW(
  TE.fromPredicate(hasActiveActivation => hasActiveActivation, identity)
),
TE.fold(
  () =>
    TE.of(
      blockedInboxOrChannel.includes(BlockedInboxOrChannelEnum.INBOX)
        ? blockedInboxOrChannel
        : [...blockedInboxOrChannel, BlockedInboxOrChannelEnum.INBOX]
    ),
  () =>
    TE.of(
      blockedInboxOrChannel.filter(
        el => el !== BlockedInboxOrChannelEnum.INBOX
      )
   )
)
```

ProcessMessage/handler.ts
BurnedMarshal: Fixed in 0b1169e

ProcessMessage/handler.ts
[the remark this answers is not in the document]
BurnedMarshal: a comment was added in 0b1169e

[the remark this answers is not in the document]
gunzip: it depends on the usage, if you only use one side returning a `Task` would be better (or using `never`).

gquadrati (approved):
LGTM

ProcessMessage/handler.ts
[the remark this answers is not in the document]
BurnedMarshal: `TaskEither` was replaced by a `Task` in 06a1b5c

package.json
BurnedMarshal: Fixed in 988385e

UpdateService/handler.ts
```diff
@@ -137,10 +140,22 @@ const updateServiceTask = (
           ...retrievedService,
           ...servicePayload,
           service_id: serviceId,
-          service_metadata: {
-            ...servicePayload.service_metadata,
-            token_name: adb2cTokenName
-          }
+          // Only Admins can change service category and custom_special_flow name
```
balanza, line 143: It's not clear to me how "being admin" is ensured in this flow

BurnedMarshal: Good point. I've tried to extend the comment in f2482cb

PR #176, opened 2021-12-10
[#IC-67]  Add Legal Message block management and create enpoint

fabriziopapi:
This PR let the CreateMessage proper manage the new data block containig the legal mail data.
It'll also add a new enpoint (CreateLegalMessage) that let invoke the standard CreateMessage impersonating another service.

#### List of Changes

Update Message model
Protect CreateMessage with new role for message with legal data
Add CreateLegalMessage function

#### Motivation and Context

In MVL project we require a service (the PEC-SERVER) to send legal message on behalf of another service

unit test
local system test (pending)

#### Types of changes

#### Checklist:


CreateLegalMessage/handler.ts
```diff
@@ -0,0 +1,152 @@
...
+      ),
+      TE.map(impersonatedService => {
+        // eslint-disable-next-line functional/immutable-data
+        rawRequest.headers["x-user-groups"] =
+          impersonatedService.user_groups || "";
+        // FIXME: const HEADER_USER_EMAIL = "x-user-email";
+        //        const HEADER_USER_SUBSCRIPTION_KEY = "x-subscription-id";
+        return impersonatedService;
+      }),
+      TE.map(ResponseSuccessJson),
```
gunzip, line 105: why not just return the new request instead of mutating the argument? It's not clear why this method should be an handler (since we call CreateMessage handler anyway) and not a simple function that transform the passed in request.

AleDore (approved):

PR #183, opened 2022-01-14
[#IC-181] Services with PENDING Activation can send messages temporarily

BurnedMarshal:
#### List of Changes

Add a new configurable ENV variables to set `PENDING` Service Activation grace time.
Messages sent within grace time from a `PENDING` Service Activation can be sent to the user, outside the grace period the message is discarded like for `INACTIVE` or missing Service Activation.
Update unit tests.

#### Motivation and Context

When a service has an asynchronous activation process, could happen that something goes wrong and a notification message should be sent to inform the user. Now asynchronous service activation process can set an Activation to `PENDING` at the beginning and send a message to the user when the error come; after a grace period when PENDING Service Activation are considered as ACTIVE, all the other messages will be discarded.
This functionality is needed by pagopa/io-functions-cgn#62

unit test

#### Types of changes

- [X] Chore (nothing changes by a user perspective)


utils/config.ts
```diff
@@ -114,6 +119,18 @@ export const envConfig = {
     ),
     E.toUnion
   ),
+  PENDING_ACTIVATION_GRACE_PERIOD_SECONDS: pipe(
+    E.fromNullable(DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS)(
+      process.env.PENDING_ACTIVATION_GRACE_PERIOD_SECONDS
+    ),
+    E.chain(_ =>
+      pipe(
+        NumberFromString.decode(_),
```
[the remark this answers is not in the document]
BurnedMarshal: Fixed in 46017b7

ProcessMessage/__tests__/handler.test.ts
```diff
@@ -211,33 +214,46 @@ const anActiveActivation: RetrievedActivation = {
...
+
+let clock: any;
+const ExecutionDateContext = Date.now();
 beforeEach(() => {
   // we usually prefer clearAllMocks, but tests scenarios are somehow entangled
   //  we should refactor them to have them independent, however for now we keep the workaround
   jest.resetAllMocks();
   mockRetrieveProcessingMessageData.mockImplementation(() =>
     TE.of(O.some(aCommonMessageData))
   );
```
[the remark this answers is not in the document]
BurnedMarshal: no we don't, removed in 46017b7

ProcessMessage/index.ts
```diff
@@ -79,6 +80,7 @@ const activityFunctionHandler: AzureFunction = getProcessMessageHandler({
   lProfileModel: profileModel,
   lServicePreferencesModel: servicePreferencesModel,
   optOutEmailSwitchDate: config.OPT_OUT_EMAIL_SWITCH_DATE,
+  pendingActivationGracePeriod: config.PENDING_ACTIVATION_GRACE_PERIOD_SECONDS as Second,
```
AleDore, line 83: not sure we need this explicit cast

BurnedMarshal: Good catch, the cast is needed but is wrong, because the casting is not enough. To convert a number to a `Second` type we need even the `kind` prop.
This line should be changed in
```suggestion
  pendingActivationGracePeriod: Object.assign(
          DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS,
          { kind: "Second" as const }
        ),
```
What do you think about?

AleDore: It should work but a test could help us in this case :D

AleDore: why only DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS ?

BurnedMarshal: Sorry wrong code paste, replace `DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS` with `DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS`. The unit test already present works correctly passing the Second object.

AleDore (approved):

ProcessMessage/index.ts
BurnedMarshal: Fixed in 4f8aee6

AleDore (approved):
LGTM

ProcessMessage/index.ts
gquadrati: Not sure about it, see [this post](https://medium.com/@KevinBGreene/surviving-the-typescript-ecosystem-branding-and-type-tagging-6cf6e516523d) about Branded Types: they only works at compile time and defined using _"as"_

ProcessMessage/__tests__/handler.test.ts
```diff
@@ -402,22 +423,24 @@ describe("getprocessMessageHandler", () => {
...
+    ${"message sender special service exists in user service preference and Activation is PENDING"}                            | ${"SENDER_BLOCKED"}        | ${O.some(aRetrievedProfileWithManualPreferences)}                                                             | ${O.some(anEnabledServicePreference)}                            | ${O.some(aPendingActivation)}                                                                                                          | ${aCreatedMessageEvent} | ${aSpecialMessageData}
+    ${"message sender special service exists in user service preference and Activation is PENDING near grace period"}          | ${"SENDER_BLOCKED"}        | ${O.some(aRetrievedProfileWithManualPreferences)}                                                             | ${O.some(anEnabledServicePreference)}                            | ${O.some({ ...aPendingActivation, _ts: subSeconds(ExecutionDateContext, DEFAULT_PENDING_ACTIVATION_GRACE_PERIOD_SECONDS).getTime() })} | ${aCreatedMessageEvent} | ${aSpecialMessageData}
```
gquadrati, line 443: how can this scenario be possible?

gquadrati (approved):
LGTM

ProcessMessage/index.ts
BurnedMarshal: Good point, but i think that `Second` is not a _Branded Type_ [link](https://github.com/pagopa/io-ts-commons/blob/master/src/units.ts)

gquadrati: I think it is exactly a Branded Type, even if we called it `kind` and not `__brand`

ProcessMessage/__tests__/handler.test.ts
gquadrati: Ok, I thought that service preferences were bounded to special service activation, my bad

ProcessMessage/index.ts
BurnedMarshal: Restored casting in 77022a0

PR #186, opened 2022-01-20
track "processed" status

lucacavallaro:
#### List of Changes

I added a `trackEvent` in `ProcessMessage` function that tracks the `PROCESSED` message status with also the time elapsed since the message has been created.

#### Motivation and Context

This change allows us to track the time elapsed from creation to processing of the message. This is an important metric to evaluate when developing new functionality on IO.

**This change has NOT been tested** as I am unable to setup the local environment. Please help me test this change!

#### Types of changes

#### Checklist:

### DIFF
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -513,6 +513,18 @@ export const getProcessMessageHandler = ({
             createdMessageEvent.message.id
           )(MessageStatusValueEnum.PROCESSED)();
 
+          telemetryClient.trackEvent({
+            name: "api.messages.processed",
+            properties: {
+              fiscalCode: toHash(profile.fiscalCode),
+              messageId: createdMessageEvent.message.id,
+              mode: profile.servicePreferencesSettings.mode,
+              senderId: createdMessageEvent.message.senderServiceId,
+              timeFromCreated: Date.now() - newMessageWithoutContent.createdAt.getTime()
+            },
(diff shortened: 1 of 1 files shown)


ProcessMessage/handler.ts
```diff
@@ -513,6 +513,18 @@ export const getProcessMessageHandler = ({
             createdMessageEvent.message.id
           )(MessageStatusValueEnum.PROCESSED)();
 
+          telemetryClient.trackEvent({
+            name: "api.messages.processed",
+            properties: {
+              fiscalCode: toHash(profile.fiscalCode),
```
gquadrati, line 519: not sure if fiscalCode could be useful in this scenario

ProcessMessage/handler.ts
```diff
@@ -513,6 +513,18 @@ export const getProcessMessageHandler = ({
...
+              messageId: createdMessageEvent.message.id,
+              mode: profile.servicePreferencesSettings.mode,
+              senderId: createdMessageEvent.message.senderServiceId,
+              timeFromCreated: Date.now() - newMessageWithoutContent.createdAt.getTime()
+            },
+            tagOverrides: { samplingEnabled: "false" }
```
gquadrati, line 525: Since we'll treat this event for statistical analysis, maybe disabling sampling it's not necessary (but I see it's disabled in creation event too)

gquadrati (commented):
I see no problem, just small comments.
@balanza can you take a look?

PR #212, opened 2022-06-22
[#ICC-103] Add payment-status to message details

fabriziopapi:
Add message status as boolean to message details returned to Organizations. This info will be retrieved directly from payment-updater.

#### List of Changes

add payment status to message details
add payment updater call

#### Motivation and Context

The Organization subscribed to premium services could retrieved payment status for advanced messages.

unit tests

#### Types of changes

#### Checklist:


GetMessage/handler.ts
```diff
@@ -306,6 +319,35 @@ export function GetMessageHandler(
     }
     const maybeMessageStatus = errorOrMaybeMessageStatus.right;
 
+    const errorOrPaymentStatus = await pipe(
```
AleDore, line 322: We must call this API only if:
- The message is ADVANCED
- The caller has the correct permission to see this info
- The message has been processed (pending=false) in order to avoid this call while checking for processing status (i.e: polling)

PR #214, opened 2022-07-04
[#ICC-103] Add payment-status from payment-updater to message details

fabriziopapi:
Add message status as boolean to message details returned to Organizations. This info will be retrieved directly from payment-updater.

#### List of Changes

add payment status to message details
add payment updater call

#### Motivation and Context

The Organization subscribed to premium services could retrieved payment status for advanced messages.

unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/GetMessage/handler.ts
+++ b/GetMessage/handler.ts
@@ -93,7 +93,7 @@ import { FeatureLevelTypeEnum } from "../generated/definitions/FeatureLevelType"
 
 import { PaymentUpdaterClient } from "../clients/payment-updater";
 import { errorsToError } from "../utils/responses";
-import { Payment } from "../generated/payment-updater/Payment";
+import { ApiPaymentMessage } from "../generated/payment-updater/ApiPaymentMessage";
 import { PaymentStatusEnum } from "../generated/definitions/PaymentStatus";
 import { MessageReadStatusAuth } from "./userPreferenceChecker/messageReadStatusAuth";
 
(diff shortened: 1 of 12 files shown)


GetMessage/handler.ts
```diff
@@ -164,6 +170,78 @@ export const getReadStatusForService = (
...
+});
+const eligibleForPaymentStatus = (
+  messageWithContent: unknown
+): E.Either<t.Errors, unknown> =>
+  pipe(
+    messageWithContent,
+    E.right,
+    E.chainFirst(mwc => WithPayment.decode(mwc)),
+    E.chainFirst(mwc => WithStatusProcessed.decode(mwc))
+  );
```
Garma00, line 194: This code can be refactored as follow ?
```ts 
pipe(
  WithPayment.decode(messageWithContent),
  E.chainFirst(WithStatusProcessed.decode)
);
```

Garma00, line 194: Can this code can be refactored as follow? 
```ts
pipe(
  WithPayment.decode(messageWithContent),
  E.chainFirst(mwc => WithStatusProcessed.decode(mwc))
);
```

fabriziopapi: You could, but in that way the output tipo wille be a `WithPayment` instead of `typeof messageWithContent`.
Using chainFirst for both the decode we only impact the output of left path.

gquadrati (approved):
LGTM

PR #215, opened 2022-07-05
[#IOCIT-36] Override sender_allowed for Special Services with Activation

BurnedMarshal:
#### List of Changes

`GetLimitedProfile` and `GetLimitedProfileByPost` returns `sender_allowed` property based on `Activation` for special services.

#### Motivation and Context

A special service can use `GetLimitedProfile` to verify the presence of the user into the App IO and the current activation status with only one API call. Nevertheless now the `server_allowed` information is coherent with the status.

Unit test

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

#### Checklist:

- [x] Extend the algorithm for the `PENDING` activation status use case.

### DIFF
--- a/ProcessMessage/handler.ts
+++ b/ProcessMessage/handler.ts
@@ -3,7 +3,19 @@
 import { Context } from "@azure/functions";
 import { BlockedInboxOrChannelEnum } from "@pagopa/io-functions-commons/dist/generated/definitions/BlockedInboxOrChannel";
 import { EUCovidCert } from "@pagopa/io-functions-commons/dist/generated/definitions/EUCovidCert";
+import { FiscalCode } from "@pagopa/io-functions-commons/dist/generated/definitions/FiscalCode";
+import { MessageStatusValueEnum } from "@pagopa/io-functions-commons/dist/generated/definitions/MessageStatusValue";
+import { PaymentDataWithRequiredPayee } from "@pagopa/io-functions-commons/dist/generated/definitions/PaymentDataWithRequiredPayee";
+import {
+  ServicesPreferencesMode,
+  ServicesPreferencesModeEnum
+} from "@pagopa/io-functions-commons/dist/generated/definitions/ServicesPreferencesMode";
+import { ActivationModel } from "@pagopa/io-functions-commons/dist/src/models/activation";
 import { MessageModel } from "@pagopa/io-functions-commons/dist/src/models/message";
+import {
+  getMessageStatusUpdater,
+  MessageStatusModel
+} from "@pagopa/io-functions-commons/dist/src/models/message_status";
 import { ProfileModel } from "@pagopa/io-functions-commons/dist/src/models/profile";
 import {
   makeServicesPreferencesDocumentId,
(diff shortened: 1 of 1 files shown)


utils/__tests__/profile.test.ts
```diff
@@ -258,14 +285,109 @@ describe("getLimitedProfileTask", () => {
...
+        false,
+        [],
+        mockServicePreferenceModel,
+        mockServiceActivationModel,
+        canSendMessageOnActivation,
+        mockTelemetryClient
+      )();
+      result.apply(mockExpressResponse);
+
+      expect(mockServicePreferenceModel.find).not.toHaveBeenCalled();
```
michaeldisaro, line 346: I'd check that `mockServiceActivationModel` has been called with something. In normal services tests we checked that it was not called, for special services maybe we can check that it is actually called.

michaeldisaro (changes requested):
Just some minor changes

utils/profile.ts
```diff
@@ -235,52 +242,82 @@ Task<IGetLimitedProfileResponses> =>
...
-          })
+    TE.chain(profile =>
+      pipe(
+        O.fromNullable(userAttributes.service.serviceMetadata),
+        O.map(serviceMetadata => serviceMetadata.category),
+        TE.fromPredicate(
+          category =>
+            O.isSome(category) &&
+            category.value === SpecialServiceCategoryEnum.SPECIAL,
+          identity
```
fabriziopapi, line 253:
```suggestion
               profile,
        TE.fromPredicate(
          () =>
            userAttributes.service.serviceMetadata?.category ===
            SpecialServiceCategoryEnum.SPECIAL,
          identity
```
Could we simplify it with something like this?

michaeldisaro: I tried it yesterday on another PR too, but I ended thinking that passing from an Option, like @BurnedMarshal did, gives a more precise idea of what we want to obtain. The second one, even if more simple and concise is not declaring an intent imho... it seems more "imperative" to me. But I think it's @BurnedMarshal choice...

utils/__tests__/profile.test.ts
BurnedMarshal: fixed in [dec75db](https://github.com/pagopa/io-functions-services/pull/215/commits/dec75dbd5fcfb899e0321cd85c761edf3982c68a)

utils/__tests__/profile.test.ts
```diff
@@ -258,14 +285,109 @@ describe("getLimitedProfileTask", () => {
...
+
+      expect(result.kind).toBe("IResponseSuccessJson");
+
+      expect(mockExpressResponse.json).toHaveBeenCalledWith(
+        expect.objectContaining({ sender_allowed: expected })
+      );
+    }
+  );
+
+  it("should responde with an ResponseErrorInternal if an error occours accessing the activation", async () => {
```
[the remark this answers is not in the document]
BurnedMarshal: Fixed in [c16cec7](https://github.com/pagopa/io-functions-services/pull/215/commits/c16cec709c959822eca46a98a08fb154a9daaafa)

michaeldisaro (approved):

utils/services.ts
```diff
@@ -30,3 +34,29 @@ export const authorizedForSpecialServicesTask = (
...
+ * Configure a function with grace period that
+ * returns true if:
+ * - the Service Activation is present and ACTIVE
+ * - the Service Activation is present and PENDING within the grace period
+ *
+ * @param pendingActivationGracePeriod PENDING grace period in seconds
+ */
+export const canSendMessageOnActivationWithGrace = (
+  pendingActivationGracePeriod: Second
+) => (maybeActivation: O.Option<RetrievedActivation>): boolean =>
```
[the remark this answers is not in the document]
BurnedMarshal: Fixed in [12fd0ea](https://github.com/pagopa/io-functions-services/pull/215/commits/12fd0ea8d1c8800eecb08f2753cdc173844b5b80)

PR #218, opened 2022-07-15
[#ICC-150]  Add feature flag FF_PAYMENT_STATUS_ENABLED

gquadrati:
#### List of Changes

- Add FF_PAYMENT_STATUS_ENABLED, with false as default

#### Motivation and Context

This is needed for publishing #212 disabled, waiting the Payment Updater to be up and running

Unit test

#### Types of changes

#### Checklist:


GetMessage/handler.ts
```diff
@@ -435,6 +450,7 @@ export const GetMessageHandler = (
  */
 // eslint-disable-next-line max-params, prefer-arrow/prefer-arrow-functions
 export function GetMessage(
+  { FF_PAYMENT_STATUS_ENABLED }: IConfig,
```
AleDore, line 453: Why destructuring object here instead of passing a simple boolean?

utils/config.ts
```diff
@@ -75,6 +77,8 @@ export const IConfig = t.intersection([
     FF_INCOMPLETE_SERVICE_WHITELIST: CommaSeparatedListOf(ServiceId),
     FF_OPT_IN_EMAIL_ENABLED: t.boolean,
 
+    FF_PAYMENT_STATUS_ENABLED: withFallback(BooleanFromString, false),
```
AleDore, line 80: <3

GetMessage/handler.ts
gquadrati: Just in case we'll need more values in future, nothing else actually.. What do you prefer?

AleDore: Since we have to remove it once we have a stable backend service, It's ok :D

AleDore (approved):

GetMessage/handler.ts
gquadrati: Right :D

PR #230, opened 2022-10-27
[ICC-201] Set ttl of message and message-status to 3 years for not registered users

Garma00:
#### List of Changes

* Added logic to set `ttl` to 3 years when user is not found;
* Added unit test;

#### Motivation and Context

In order to remove messages sent to users not registered we want to set the 
`ttl` to 3 years to `message` and `message-status`.

Unit test

#### Types of changes

#### Checklist:


ProcessMessage/handler.ts
```diff
@@ -422,6 +425,23 @@ export const getProcessMessageHandler = ({
...
+                // setting TTL to 3 years for message entry
+                lMessageModel.patch(
+                  [
+                    newMessageWithoutContent.id,
+                    newMessageWithoutContent.fiscalCode
+                  ],
+                  { ttl: 94670856 } as Partial<MessageWithoutContent>
+                )
+              ),
               TE.getOrElse(e => {
```
fabriziopapi, line 445: This error managmenet is not enough: if the update ttl or the patch fails, it will simple throw a not verbose error.
I think we need a specific ai track to specific manage those cases.

ProcessMessage/handler.ts
```diff
@@ -422,6 +425,23 @@ export const getProcessMessageHandler = ({
...
+                )
+              ),
+              TE.chain(() =>
+                // setting TTL to 3 years for message entry
+                lMessageModel.patch(
+                  [
+                    newMessageWithoutContent.id,
+                    newMessageWithoutContent.fiscalCode
+                  ],
+                  { ttl: 94670856 } as Partial<MessageWithoutContent>
```
gquadrati, line 442: I'd rather put ttl value into a config variable.

ProcessMessage/handler.ts
```diff
@@ -422,6 +432,55 @@ export const getProcessMessageHandler = ({
...
+                pipe(
+                  // setting TTL to 3 years for message entry
+                  lMessageModel.patch(
+                    [
+                      newMessageWithoutContent.id,
+                      newMessageWithoutContent.fiscalCode
+                    ],
+                    { ttl: TTL_FOR_USER_NOT_FOUND } as Partial<
+                      MessageWithoutContent
+                    >
```
gquadrati, line 467: We know that `ttl` is not allowed by patch, even if it's correctly working at runtime. Maybe it's better to add a little comment to explain why we need this cast.

ProcessMessage/handler.ts
```diff
@@ -422,6 +432,55 @@ export const getProcessMessageHandler = ({
...
+                  ),
+                  TE.mapLeft((error: CosmosErrors) => {
+                    telemetryClient.trackEvent({
+                      name: "api.messages.create.failttlset",
+                      properties: {
+                        error: `Something went wrong trying to update the ttl for the message | ${error.kind}`,
```
gquadrati, line 473: Ditto

ProcessMessage/handler.ts
```diff
@@ -422,6 +432,55 @@ export const getProcessMessageHandler = ({
...
+                      newMessageWithoutContent.id,
+                      newMessageWithoutContent.fiscalCode
+                    ],
+                    { ttl: TTL_FOR_USER_NOT_FOUND } as Partial<
+                      MessageWithoutContent
+                    >
+                  ),
+                  TE.mapLeft((error: CosmosErrors) => {
+                    telemetryClient.trackEvent({
+                      name: "api.messages.create.failttlset",
```
gquadrati, line 471:
```suggestion
                      name: "api.messages.create.fail-message-ttl-set",
```

ProcessMessage/handler.ts
```diff
@@ -422,6 +432,55 @@ export const getProcessMessageHandler = ({
...
+              TE.chain(() =>
+                pipe(
+                  // setting TTL to 3 years for message-status entries
+                  lMessageStatusModel.updateTTLForAllVersions(
+                    [newMessageWithoutContent.id],
+                    TTL_FOR_USER_NOT_FOUND
+                  ),
+                  TE.mapLeft((error: CosmosErrors) => {
+                    telemetryClient.trackEvent({
+                      name: "api.messages.create.failstatusttlset",
```
gquadrati, line 444: A little more readable

```suggestion
                      name: "api.messages.create.fail-status-ttl-set",
```

ProcessMessage/__tests__/handler.test.ts
```diff
@@ -743,6 +768,65 @@ describe("getprocessMessageHandler", () => {
         skipPreferenceMock ? 0 : 1
       );
       expect(activationFindLastVersionMock).toBeCalledTimes(0);
+      expect(patchMessageMock).not.toHaveBeenCalled();
+      expect(updateTTLForAllVersionsMock).not.toHaveBeenCalled();
     }
   );
+
+  it("it should fail if profile is not found, and the ttl must be defined", async () => {
```
gquadrati, line 776:
```suggestion
  it("it should end with rejection if profile is not found, and the ttl must be defined", async () => {
```

gquadrati (changes requested):
We also need to introduce this change under feature flag.

__integrations__/environments/env.io-functions-services
```diff
@@ -53,6 +53,8 @@ APIM_SUBSCRIPTION_KEY="NonEmptyString"
 # Min App Version Handling Read Access
 # ----------
 
+TTL_FOR_USER_NOT_FOUND=94670856 
+FEATURE_FLAG=all
```
gquadrati, line 57: I was expecting integration tests to fail. Please check if it's really calling new code.

gquadrati (changes requested):

ProcessMessage/handler.ts
```diff
@@ -347,7 +351,13 @@ const createMessageOrThrow = async (
   }
 };
 
+// ttl can only be a positive integer or -1
+const ttlType = t.union([NonNegativeInteger, t.literal(-1)]);
+type TtlType = t.TypeOf<typeof ttlType>;
+
 export interface IProcessMessageHandlerInput {
+  readonly TTL_FOR_USER_NOT_FOUND: TtlType;
+  readonly isUserForFeatureFlag: (fc: FiscalCode) => boolean;
```
gquadrati: Please, rename it to `isUserEligibleForNewFeature` everywhere

utils/featureFlags.ts
```diff
@@ -0,0 +1,28 @@
+import * as t from "io-ts";
+
+export const FeatureFlag = t.union([
+  t.literal("all"),
+  t.literal("beta"),
+  t.literal("canary"),
+  t.literal("none")
+]);
+export type FeatureFlag = t.TypeOf<typeof FeatureFlag>;
```
gquadrati, line 9: Please, move to `enumType` like [here](https://github.com/pagopa/io-ts-commons/blob/3ec0e9bdf706ac22d02853c390e8317392299617/src/environment.ts#L12).
This way we can avoid fixed string and use enum value instead.

gquadrati (changes requested):

gquadrati (approved):

PR #239, opened 2022-11-23
[#IOPAE-77] Fix how handle sandboxFiscalCode

LoFrance:
This PR add `sandbox fiscal code` if we don't pass any `authorized_recipients` inside body payload or add it to the list of `authorized_recipients`.

#### Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/CreateService/handler.ts
+++ b/CreateService/handler.ts
@@ -156,10 +156,12 @@ export function CreateServiceHandler(
   apiClient: APIClient,
   generateObjectId: ObjectIdGenerator,
   productName: NonEmptyString,
-  authorizedRecipients: NonEmptyString | ReadonlyArray<FiscalCode>
+  sandboxFiscalCode: NonEmptyString
 ): ICreateServiceHandler {
   // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
   return (context, __, ___, userAttributes, servicePayload) => {
+    const authorizedRecipients = (servicePayload as Service)
+      .authorized_recipients;
     const subscriptionId = generateObjectId();
     context.log.info(
       `${logPrefix}| Creating new service with subscriptionId=${subscriptionId}`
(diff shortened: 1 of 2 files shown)


iwoak:
@LoFrance lint it ;)

CreateService/handler.ts
```diff
@@ -185,10 +204,12 @@ export function CreateServiceHandler(
...
-          // (sandboxFiscalCode as unknown) as FiscalCode,
-          Array.isArray(authorizedRecipients)
-            ? authorizedRecipients
-            : [authorizedRecipients],
+          O.isSome(maybeAuthorizedRecipients)
+            ? [
+                (sandboxFiscalCode as unknown) as FiscalCode,
+                ...maybeAuthorizedRecipients.value
+              ]
+            : [(sandboxFiscalCode as unknown) as FiscalCode],
```
balanza, line 212: Alternatively:
```ts
[
  (sandboxFiscalCode as unknown) as FiscalCode,
   ...O.getOrElse([])
]
```

CreateService/handler.ts
```diff
@@ -147,6 +148,21 @@ const createServiceTask = (
     200
   );
 
+export const getAuthorizedRecipientsFromPayload = (
+  servicePayload: ServicePayload
+): O.Option<null | ReadonlyArray<FiscalCode>> =>
```
balanza, line 153: Why `null`? `Option`s are used to avoid null-ish values, if we may encounter a `null` maybe we are doing it wrong.

balanza (changes requested):
Please avoid `Option<null>` objects as it leads to inconsistent state.

LoFrance:
> Please avoid `Option<null>` objects as it leads to inconsistent state.

Yep, sorry it was a mistake at the beginning of writing the function's signature and I forget to remove it.

CreateService/handler.ts
LoFrance: Fix with cecbe9f

CreateService/handler.ts
LoFrance: Done with cecbe9f

iwoak (approved):
LGTM

balanza (approved):

PR #287, opened 2023-11-09
[#IOPAE-710] Fix Get and Update service cms openAPI

StefanoDoc:
there is a typo in the open api that says that you can call the cms `retrieveService` with the service key and that is not true plus the `updateService` can return a `409 error` for no transition matched when you call `updateService` on a submitted service
#### Types of changes

#### Checklist:


giuseppedipinto (approved):
LGTM

PR #300, opened 2024-03-01
Iocom-1155

michaeldisaro:
#### List of Changes
Added error when no configuration id is sent.
Added rc configuration retrieval from functions if an id is sent.
Added various checks on RC configuration, the most important are that it must exist and it must be owned from the subscription's user.

#### Motivation and Context
When Remote Content data (third_party_data) is sent with a new message IT MUST contain the new configuration_id field that MUST point to an owned valid remote content configuration that is mandatory to have working remote contents. 

build
lint
unit tests

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [x] My change requires a change to the documentation.


CreateMessage/handler.ts
```diff
@@ -307,6 +347,61 @@ export function CreateMessageHandler(
...
+          // check the ownership of the remote content configuration if sent with third_party_data
+          if (
+            auth.userId !==
+            getRCConfigurationResponseOrError.right.value.user_id
+          ) {
+            return ResponseErrorForbiddenNotYourConfiguration;
+          }
+          break;
+        default:
+          break;
```
Garma00, line 401: Shall we return 500 here?

CreateMessage/__tests__/handler.test.ts
```diff
@@ -538,12 +588,52 @@ describe("CreateMessageHandler", () => {
...
       some(anotherFiscalCode)
     );
 
-    expect(r.kind).toBe("IResponseSuccessRedirectToResource");
+    expect(r.kind).toBe("IResponseErrorForbiddenNotYourConfiguration");
+  });
+
+  it("should return 403 ok if a configuration_id is not provided", async () => {
+    const mockGenerateObjId = jest
+      .fn()
```
Garma00, line 605: Can we add some tests in case the `mockMessagesServicesApiClient` return `400` `404` and `500`?

CreateMessage/handler.ts
michaeldisaro: Nice catch... Maybe yes.

CreateMessage/__tests__/handler.test.ts
michaeldisaro: https://github.com/pagopa/io-functions-services/pull/300/commits/c2033fe4ab849188ef9ddcab44f07affb46826a8

CreateMessage/handler.ts
```diff
@@ -259,28 +298,43 @@ const redirectToNewMessage = (
...
+ ** resouce link: https://docs.microsoft.com/en-us/rest/api/apimanagement/current-ga/subscription/get
+ */
+export const parseOwnerIdFullPath = (
+  fullPath: NonEmptyString
+): NonEmptyString =>
+  pipe(
+    fullPath,
+    f => f.split("/"),
+    a => a[a.length - 1] as NonEmptyString
+  );
```
lucacavallaro, line 313: 1. what if `a[index]` does not exists?
2. we should avoid to use `as`, the only way to make sure that the return type is `NonEmptyString` we have to parse with `NonEmptyString.decode`

CreateMessage/handler.ts
```diff
@@ -307,6 +361,63 @@ export function CreateMessageHandler(
     const { service, email: serviceUserEmail } = userAttributes;
     const { authorizedRecipients, serviceId } = service;
 
+    // Check configuration id if third_party_data has been sent
+    const isRemoteContentMessage = !!messagePayload.content.third_party_data;
+    const hasRemoteContentMessageConfigurationId = !!messagePayload.content
```
lucacavallaro, line 366: it would be better to parse the `messagePayload` to an appropriate "message variant" (`RemotedContentMessage`, with a dedicated codec) to centralize this constraint

CreateMessage/handler.ts
```diff
@@ -307,6 +361,63 @@ export function CreateMessageHandler(
...
+          );
+        case 500:
+          return ResponseErrorInternal(
+            "Generic error happened while trying to get remote content configuration."
+          );
+        case 200:
+          // check the ownership of the remote content configuration if sent with third_party_data
+          if (
+            parseOwnerIdFullPath(auth.userId) !==
+            getRCConfigurationResponseOrError.right.value.user_id
```
lucacavallaro, line 409: seems like that there are multiple names that refers to the same thing (`ownerId`, `userId`)

lucacavallaro (changes requested):

CreateMessage/handler.ts
Garma00: For the first point:
We are using the `split` method on a `NonEmptyString` so i think that the scenario of a non existing `a[index]` will never happen because the split method should always return an array with at least one element.
We can add a check on the length before the split to be sure tho

michaeldisaro: As garma said, split always return an array. In fact that code is working also when we pass a simple user id without full path. Splitting by slash gives us an array with an only word inside.

CreateMessage/handler.ts
michaeldisaro: where is `ownerId`?

CreateMessage/handler.ts
michaeldisaro: this would be particularly difficult and out of scope in this moment. Since we will have to migrate to a v2 version in the next future, I'd keep this kind of optimization in that job.

Garma00 (approved):

PR #308, opened 2024-04-24
[IOCOM-1313] feat(openapi): Added remote content manage api

michaeldisaro:
#### List of Changes
Added remote content manage APIs and definitions.

#### Motivation and Context
We need to expose these new APIs on the public guide.

#### Types of changes

### DIFF
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -1052,7 +1052,7 @@ paths:
           description: Internal server error
           schema:
             $ref: '#/definitions/ProblemJson'
-  /service-messages/api/v1/remote-contents/configurations:
+  /service-messages/manage/api/v1/remote-contents/configurations:
     post:
       tags:
         - remote-content
@@ -1203,7 +1203,7 @@ paths:
         - application/json
       produces:
         - application/json
-  /service-messages/api/v1/remote-contents/configurations/{configurationId}:
+  /service-messages/manage/api/v1/remote-contents/configurations/{configurationId}:
     get:
       tags:
         - remote-content
--- a/openapi/index.yaml
+++ b/openapi/index.yaml
@@ -98,11 +98,11 @@ info:
       1. (optional) Call [getMessage](#operation/getMessage) to check whether
          the message has been notified to the recipient.
 host: api.io.pagopa.it
-basePath: /api/v1
+basePath: /
 schemes:
   - https
 paths:
-  /messages:
+  /api/v1/messages:
     post:
       operationId: submitMessageforUserWithFiscalCodeInBody
       summary: Submit a Message passing the user fiscal_code in the request body
@@ -174,7 +174,7 @@ paths:
           description: The message cannot be delivered.
           schema:
             $ref: '#/definitions/ProblemJson'
-  /messages/{fiscal_code}:
+  /api/v1/messages/{fiscal_code}:
     post:
       operationId: submitMessageforUser
       summary: Submit a Message passing the user fiscal_code as path parameter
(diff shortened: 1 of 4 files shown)


openapi/index.yaml
```diff
@@ -1052,6 +1052,296 @@ paths:
           description: Internal server error
           schema:
             $ref: '#/definitions/ProblemJson'
+  /service-messages/manage/api/v1/remote-contents/configurations:
```
iwoak, line 1055: We could have /api/v1/manage/messages/ as the path prefix for any message management API?
Could the final path for the remote content configuration API be something like this /api/v1/manage/messages/remote-content/configurations?
My point is:
- be as conservative as possible with respect to other actual APIs
- have a self-explanatory path 
Perhaps @rodoviti can help us as he has already addressed this with the services manage api

And in a future step, we'll decide how to manage different stream APIs.

iwoak: maybe it's easier to have something like /api/v1/communication/manage/remote-content/configurations

gunzip: "manage" is superfluous imho since the action is implied by the http verb which acts on the configuration resource.

iwoak: Actually I agree, we can't use /api/v1/messages as a path prefix because it matches other existing endpoints with different backends, we could use a messages synonym (e.g. /api/v1/communication/remote-content/configurations) or something else and 'manage' is similar to others (a general revision is needed to enable a better api stream ownership)

michaeldisaro: This part of the path ` /service-messages/manage` is made by 2 reasons:
1. we migrated the `service-message` part from AKS to Azure Functions and to have a 0 downtime we maintained the original openapi specs.
2. we added `manage` because we are exposing these specs to TPs and we had to distinguish from the internal use of these APIs at the APIM's level.

We could now achieve what you are saying if it is really necessary, we should delay the delivery to TPs of these APIs that was for the next week because we have now to prepare a new openapi without the ` /service-messages/manage` part for both the APIs, and then create 2 new APIs on APIM distinguishing the internal ones with a `internal` subpath, and then deploy to the Azure Functions the new functions with new paths without the `service-messages` part.

angelo-fiorillo (approved):
OK per me

openapi/index.yaml.template
```diff
@@ -994,6 +994,298 @@ paths:
...
+        - in: body
+          name: body
+          required: true
+          schema:
+            $ref: "#/definitions/NewRCConfigurationPublic"
+          x-examples:
+            text/json: |
+              "name": "Descriptive congiguration name",
+              "description": "Configuration description",
+              "has_precondition": "NONE",
```
Garma00, line 1017: `
```suggestion
              "has_precondition": "NEVER",
```

openapi/index.yaml.template
```diff
@@ -994,6 +994,298 @@ paths:
...
+        "201":
+          description: Configuration created
+          schema:
+            $ref: "#/definitions/RCConfigurationPublic"
+          examples:
+            text/json: |
+              "configuration_id": "01ARZ3NDEKTSV4RRFFQ69G5FAV",
+              "name": "Configuration name",
+              "description": "Configuration description",
+              "has_precondition": "NONE",
```
Garma00, line 1051: ditto here and in other examples

Garma00 (approved):
Just a little change

michaeldisaro:
Please do not merge. I'll do it when ready.

PR #309, opened 2024-07-15
[#IOCOM-1597] Managed LEGACY user as app UNKNOWN

michaeldisaro:
#### List of Changes
Added check to manage LEGACY user service preference as UNKNOWN like if we do not know the user's app version.

#### Motivation and Context
We need to manage users with LEGACY service preferences because actually they were not considered to exists.

- build
- unit test
- lint 

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly. **RFC + DR**

### DIFF
--- a/GetMessage/userPreferenceChecker/__tests__/messageReadStatusAuth.test.ts
+++ b/GetMessage/userPreferenceChecker/__tests__/messageReadStatusAuth.test.ts
@@ -93,7 +93,12 @@ describe("canAccessMessageReadStatus |> ok", () => {
 
   it("should return false if profile servicePreferencesSettings is of type LEGACY (version is -1)", async () => {
     mockProfileFindLastVersionByModelId.mockReturnValueOnce(
-      TE.of(O.some(aRetrievedProfile))
+      TE.of(
+        O.some({
+          ...aRetrievedProfile,
+          lastAppVersion: MIN_READ_STATUS_PREFERENCES_VERSION
+        })
+      )
     );
 
     const res = await canAccessMessageReadStatus(
@@ -268,9 +273,7 @@ describe("canAccessMessageReadStatus |> Errors", () => {
       MIN_READ_STATUS_PREFERENCES_VERSION
     )(aServiceId, aFiscalCode)();
 
-    expect(res).toStrictEqual(
-      E.right(false)
-    );
+    expect(res).toStrictEqual(E.right(false));
 
     // Do not call service preferences like if app version is UNKNOWN
     expect(mockServicePreferencesFind).not.toHaveBeenCalled();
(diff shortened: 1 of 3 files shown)


GetMessage/userPreferenceChecker/__tests__/userPreferencesCheckerFactory.test.ts
```diff
@@ -138,6 +139,24 @@ describe("userPreferencesCheckerFactory |> userPreferencesCheckerFactory", () =>
...
+    mockServicePreferencesGetter.mockReturnValueOnce(TE.of(O.none));
+
+    const res = await userPreferencesCheckerFactory(
+      {
+        ...aRetrievedProfile,
+        lastAppVersion: "UNKNOWN"
+      },
+      mockServicePreferencesGetter,
+      MIN_READ_STATUS_PREFERENCES_VERSION
+    ).canAccessMessageReadStatus(aServiceId, aFiscalCode)();
```
Garma00, line 152: This test is correct but here we are not testing the exact scenario that we want to, I would use a `MIN_READ_STATUS_PREFERENCES_VERSION` that would not make `isAppVersionHandlingReadAuthorization` pass. In fact this test would pass even if the `servicePreferenceSettings.version` would be a `NonNegativeInteger`.
This is what i suggest: 
```suggestion
  it("should call userPreferenceCheckerVersionUNKNOWNToVersionWithReadAuth if profile servicePreferencesSettings is of type LEGACY (version is -1)", async () => {
    mockServicePreferencesGetter.mockReturnValueOnce(TE.of(O.none));

    const res = await userPreferencesCheckerFactory(
      {
        ...aRetrievedProfile,
        lastAppVersion: "UNKNOWN"
      },
      mockServicePreferencesGetter,
      1.0.0 as Semver
    ).canAccessMessageReadStatus(aServiceId, aFiscalCode)();
```

michaeldisaro: You found a problem, but the analysis was wrong. Legacy should go "unknown" when app version is not unknown.
I fixed with
```
{
        ...aRetrievedProfile,
        lastAppVersion: MIN_READ_STATUS_PREFERENCES_VERSION
      },
```
Now if you send a non negative integer this will not pass.

GetMessage/userPreferenceChecker/__tests__/messageReadStatusAuth.test.ts
```diff
@@ -91,6 +91,23 @@ describe("canAccessMessageReadStatus |> ok", () => {
     expect(mockServicePreferencesFind).not.toHaveBeenCalled();
   });
 
+  it("should return false if profile servicePreferencesSettings is of type LEGACY (version is -1)", async () => {
```
Garma00, line 94: There is something with this test, it ends with success even if we pass a servicePreferenceSettings not LEGACY

michaeldisaro: fixed in https://github.com/pagopa/io-functions-services/pull/309/commits/556d47841a85139347a042991bee29c38abf78e9

Garma00 (approved):

PR #310, opened 2024-08-22
[EC-399] Azure DevOps pipelines migrated into GitHub Actions

mamu0:
#### List of Changes

The two workflows on Azure DevOps

- `code-review`
- `deploy-pipelines`

have been recreated in the form of GitHub Actions.

They have been inserted in the `.github/workflows` folder.

Added `pre-commit` configuration.

Added `infra` folder with `identity` and `repository` infrastructure, and added branch protection on `master` (already applied).

Edited `.gitignore` and `CODEOWNER` files.

#### Motivation and Context

Since all the pipelines on Azure DevOps are gradually being decommissioned, it was decided to recreate them on the different repos such as GitHub Actions.

Currently they have been tested by inserting

```
push:
  branches:
    - EC-399-devops-decommissioning
```
so as to execute them automatically at each test push.
To do the Deploy test for `deploy-pipelines` action i've used this step only to check if all will be called correctly:
```
      - name: Deploy
        run: |
          az webapp up \
            --dryrun \
            --resource-group ${{ inputs.resource_group_name }} \
            --name ${{ inputs.app_name }} \
```

To do the Deploy test for `publish-sdk` action i've executed everything except last command with effective publish.

A real deploy test is needed.

#### Types of changes

#### Checklist:

- [x] I have updated the documentation accordingly.


.github/workflows/deploy-pipelines.yaml
```diff
@@ -0,0 +1,134 @@
...
+        if: ${{ inputs.use_staging_slot == false }}
+        run: |
+          az webapp deploy \
+            --resource-group ${{ inputs.resource_group_name }} \
+            --name ${{ inputs.app_name }} \
+            --src-path  ${{ github.workspace }}/${{ env.BUNDLE_NAME }}.zip \
+            --type zip \
+            --async false
+
+      - name: Deploy to Staging Slot
```
gunzip, line 116: don't we need to set `WEBSITE_SWAP_WARMUP_PING_PATH` in order to check staging slot health before swap?

gunzip: see https://github.com/pagopa/dx/pull/77#issuecomment-2275885459

mamu0: This variable needs to be configured in the terraform infrastructure of the function, which should now be on [io-infra](https://github.com/pagopa/io-infra/blob/main/src/domains/functions/function_services.tf).

But I don't know how it should be handled in this case, in any case I don't think it should be done in this PR, @Krusty93 do you know what should be done?

gunzip: if swapping is working as is we can maintain it by now imho, and set up the needed var later in io-infra

mamu0: Regarding this the actual `swap` still needs to be tested: it needs to be merged and then verified that in a real case everything works correctly. 
For now everything has been tested in `dry-run` since the **environment** is `prod`.

Krusty93: @gunzip Correct, I think we should include that variable in function and app service modules. However, it is important that the checked path initialise the entire application and its dependencies

cc @lucacavallaro

BurnedMarshal:
Is this PR on hold? @mamu0

.github/workflows/code-review.yml
```diff
@@ -0,0 +1,64 @@
...
+    paths:
+      - "*"
+      - "!infra/*"
+      - "!.github/*"
+
+jobs:
+  js_code_review:
+    runs-on: ubuntu-22.04
+
+    steps:
```
BurnedMarshal, line 16: is missing a step to run the integration test.
Take a look in [here](https://github.com/pagopa/io-functions-services/blob/master/.devops/code-review-pipelines.yml#L81)

